PR #9, opened 2024-03-01
Publish to maven central

w4ll3:


kotlin/walletsdkrs/build.gradle
```diff
@@ -104,15 +103,14 @@ publishing {
                 password = System.getenv("GITHUB_TOKEN")
             }
         }
-    }
-    publications {
-        release(MavenPublication) {
```
sbihel, line 109: won't this removal mean the publication won't happen?

kotlin/walletsdkrs/build.gradle
```diff
@@ -104,15 +103,14 @@ publishing {
...
-            groupId = 'com.spruceid.wallet.sdk.rs'
-            artifactId = "walletsdkrs"
-            version = System.getenv("VERSION")
-
-            afterEvaluate {
-                from components.release
+        repositories {
+            maven {
+                name = "OSSRH"
+                url = "https://oss.sonatype.org/service/local/staging/deploy/maven2/"
```
sbihel, line 109: can we publish to "release" directly without having to promote manually from staging (if I'm understanding things correctly)?

sbihel (changes requested):

kotlin/walletsdkrs/build.gradle
sbihel: doesn't seem possible

sbihel (approved):

PR #22, opened 2024-07-22
Add mDoc reader functions

sbihel:
Checklist:
- [x] Figure out if the current fromatting of `ble_ident` is correct. But probably not because we go from bytes to hex string back to bytes without decoding the string.
- [x] Make the `MDocItem` type work in Kotlin, or move back to JSON as it is the type returned by `isomdl`
- [x] test on a least one platform with bluetooth
- [x] change back the local path dependency for the binary target


src/mdl_reader.rs
```diff
@@ -0,0 +1,301 @@
...
+    let uuid =
+        manager2
+            .first_central_client_uuid()
+            .ok_or_else(|| MDLReaderSessionError::Generic {
+                value: "the device did not transmit a central client uuid".to_string(),
+            })?;
+
+    // TODO this must be wrong -- it's encoded as a hex string, but then advertised as raw bytes
+    let mut ble_ident = hex::encode(ble_ident);
+    ble_ident.insert_str(0, "0x");
```
rschulman, line 102: I believe that `mdl_bridge_native` was doing it this way because it is passing the entire structure as a string through JSON and decoding it on the other side. You might be able to get away with returning a Vec<u8> here and letting UniFFI deal with serialization.

src/mdl/reader.rs
```diff
@@ -0,0 +1,285 @@
...
+    };
+    Ok(anchor)
+}
+
+// TODO what was this for? Reusing the same reader?
+#[derive(thiserror::Error, uniffi::Error, Debug)]
+pub enum MDLReaderRequestBuildError {
+    #[error("{value}")]
+    Generic { value: String },
+}
```
cobward, line 130: Either remove the comment or the enum?

src/mdl/reader.rs
```diff
@@ -0,0 +1,285 @@
...
+
+#[derive(uniffi::Enum, Debug)]
+pub enum MDocItem {
+    Text(String),
+    Bool(bool),
+    Integer(i64),
+    // Bytes(Vec<u8>), // already converted in string
+    ItemMap(HashMap<String, MDocItem>),
+    Array(Vec<MDocItem>),
+    // Image(Vec<u8>) maybe?
```
cobward, line 197: Can remove this comments?

cobward (approved):
LGTM to merge, we can always make improvements as the implementation in a real application is done and tested.

src/mdl/reader.rs
sbihel: removed the whole thing as it's unused

PR #24, opened 2024-08-01
Draft PR: VdcCollection

todd-spruceid:
Do Not Merge

This is an initial draft version of the VDC Collection.  Some work remains to be done; storage read/write calls are currently stubs, and some other comments and suggestions have yet to be incorporated.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,10 +23,9 @@ isomdl = { git = "https://github.com/spruceid/isomdl", rev = "1f4f762" }
 thiserror = "1.0.56"
 hex = "0.4.3"
 p256 = { version = "0.13.2", features = ["pkcs8"] }
-serde = "1.0.204"
+serde = { version = "1.0.204", features = ["derive"] }
 serde_cbor = "0.11.2"
 serde_json = "1.0.111"
-serde_derive = "1.0.204"
 uuid = { version = "1.6.1", features = ["v4"] }
 tracing = "0.1.40"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,9 +23,12 @@ isomdl = { git = "https://github.com/spruceid/isomdl", rev = "1f4f762" }
 thiserror = "1.0.56"
 hex = "0.4.3"
 p256 = { version = "0.13.2", features = ["pkcs8"] }
+serde = "1.0.204"
 serde_cbor = "0.11.2"
 serde_json = "1.0.111"
+serde_derive = "1.0.204"
 uuid = { version = "1.6.1", features = ["v4"] }
+tracing = "0.1.40"
 
 [dev-dependencies]
 uniffi = { version = "0.27.1", features = ["bindgen-tests"] }
--- a/src/storage_manager.rs
+++ b/src/storage_manager.rs
@@ -1,10 +1,10 @@
 use thiserror::Error;
 
 uniffi::custom_newtype!(Key, String);
-pub struct Key(String);
+pub struct Key(pub String);
 
-uniffi::custom_newtype!(Value, String);
-pub struct Value(String);
+uniffi::custom_newtype!(Value, Vec<u8>);
+pub struct Value(pub Vec<u8>);
 
 /// Enum: StorageManagerError
 ///
@@ -56,7 +56,12 @@ pub trait StorageManagerInterface: Send + Sync {
     ///
     /// Callback function pointer to native (kotlin/swift) code for
     /// getting a key.
-    fn get(&self, key: Key) -> Result<Option<Value>, StorageManagerError>;
+    fn get(&self, key: Key) -> Result<Value, StorageManagerError>;
+
+    /// Function: list
+    ///
+    /// Callback function pointer for listing available keys.
+    fn list(&self) -> Vec<Key>;
 
     /// Function: remove
     ///
--- a/src/storage_manager.rs
+++ b/src/storage_manager.rs
@@ -28,6 +28,10 @@ pub enum StorageManagerError {
     /// error is raised when that key could not be created.
     #[error("Could not make storage encryption key")]
     CouldNotMakeKey,
+
+    /// An internal problem occurred in the storage manager.
+    #[error("Internal Error")]
+    InternalError(String),
 }
 
 /// Interface: StorageManagerInterface
--- a/src/storage_manager.rs
+++ b/src/storage_manager.rs
@@ -60,12 +60,12 @@ pub trait StorageManagerInterface: Send + Sync {
     ///
     /// Callback function pointer to native (kotlin/swift) code for
     /// getting a key.
-    fn get(&self, key: Key) -> Result<Value, StorageManagerError>;
+    fn get(&self, key: Key) -> Result<Option<Value>, StorageManagerError>;
 
     /// Function: list
     ///
     /// Callback function pointer for listing available keys.
-    fn list(&self) -> Vec<Key>;
+    fn list(&self) -> Result<Vec<Key>, StorageManagerError>;
 
     /// Function: remove
     ///
--- a/src/storage_manager.rs
+++ b/src/storage_manager.rs
@@ -31,7 +31,7 @@ pub enum StorageManagerError {
 
     /// An internal problem occurred in the storage manager.
     #[error("Internal Error")]
-    InternalError(String),
+    InternalError,
 }
 
 /// Interface: StorageManagerInterface
--- a/src/vdc_collection.rs
+++ b/src/vdc_collection.rs
@@ -74,6 +74,10 @@ pub enum VdcCollectionError {
     /// Attempting to read the credential from storage failed.
     #[error("Failed to Read from Storage")]
     LoadFailed(StorageManagerError),
+
+    /// Attempting to delete a credential from storage failed.
+    #[error("Failed to Delete from Storage")]
+    DeleteFailed(StorageManagerError),
 }
 
 impl VdcCollection {
@@ -111,6 +115,14 @@ impl VdcCollection {
         match serde_cbor::de::from_slice(&raw) {
             Ok(x) => Ok(x),
             Err(_) => Err(VdcCollectionError::DeserializeFailed),
+        }
+    }
+
+    /// Remove a credential from the store.
+    pub fn delete(&self, id: &str) -> Result<(), VdcCollectionError> {
+        match self.storage.remove(self.str_to_key(id)) {
+            Ok(_) => Ok(()),
+            Err(e) => Err(VdcCollectionError::DeleteFailed(e)),
         }
     }
 
@@ -168,5 +180,93 @@ impl VdcCollection {
                 }
             }
         });
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::local_store::*;
+    use uuid::uuid;
+
+    #[test]
+    fn test_vdc() {
+        let smi = LocalStore;
+        let vdc = VdcCollection::new(Box::new(smi));
+        let payload_1: Vec<u8> = "Some random collection of bytes. ⚛".into();
+        let payload_2: Vec<u8> = "Some other random collection of bytes. 📯".into();
+        let payload_3: Vec<u8> = "Some third random collection of bytes. λ".into();
+
+        match vdc.add(
+            uuid!("00000000-0000-0000-0000-000000000001"),
+            CredentialFormat::MsoMdoc,
+            CredentialType::Iso18013_5_1mDl,
+            payload_1.clone(),
+        ) {
+            Ok(_) => {}
+            Err(_) => panic!("Unable to add the first value."),
+        }
+
+        match vdc.add(
+            uuid!("00000000-0000-0000-0000-000000000002"),
+            CredentialFormat::MsoMdoc,
+            CredentialType::Iso18013_5_1mDl,
+            payload_2.clone(),
+        ) {
+            Ok(_) => {}
+            Err(_) => panic!("Unable to add the second value."),
+        }
+
+        match vdc.add(
+            uuid!("00000000-0000-0000-0000-000000000003"),
+            CredentialFormat::MsoMdoc,
+            CredentialType::Iso18013_5_1mDl,
+            payload_3.clone(),
+        ) {
+            Ok(_) => {}
+            Err(_) => panic!("Unable to add the third value."),
+        }
+
+        match vdc.get("00000000-0000-0000-0000-000000000002") {
+            Ok(x) => assert!(x.payload == payload_2.clone()),
+            Err(_) => panic!("Failed to get the second value"),
+        }
+
+        match vdc.get("00000000-0000-0000-0000-000000000001") {
+            Ok(x) => assert!(x.payload == payload_1.clone()),
+            Err(_) => panic!("Failed to get the first value"),
+        }
+
+        match vdc.get("00000000-0000-0000-0000-000000000003") {
+            Ok(x) => assert!(x.payload == payload_3.clone()),
+            Err(_) => panic!("Failed to get the third value"),
+        }
+
+        let list = vdc.all_entries();
+
+        assert!(list.len() == 3);
+
+        match vdc.delete("00000000-0000-0000-0000-000000000002") {
+            Ok(_) => {}
+            Err(_) => panic!("Failed to delete the second value."),
+        }
+
+        let list = vdc.all_entries();
+
+        assert!(list.len() == 2);
+
+        match vdc.delete("00000000-0000-0000-0000-000000000001") {
+            Ok(_) => {}
+            Err(_) => panic!("Failed to delete the first value."),
+        }
+
+        match vdc.delete("00000000-0000-0000-0000-000000000003") {
+            Ok(_) => {}
+            Err(_) => panic!("Failed to delete the third value."),
+        }
+
+        let list = vdc.all_entries();
+
+        assert!(list.len() == 0);
     }
 }
--- a/src/vdc_collection.rs
+++ b/src/vdc_collection.rs
@@ -74,12 +74,10 @@ impl VdcCollection {
         ctype: CredentialType,
         payload: Vec<u8>,
     ) -> Result<(), VdcCollectionError> {
-        let val;
-
-        match serde_cbor::to_vec(&Credential::new(id, format, ctype, payload)) {
-            Ok(x) => val = x,
+        let val = match serde_cbor::to_vec(&Credential::new(id, format, ctype, payload)) {
+            Ok(x) => x,
             Err(_) => return Err(VdcCollectionError::SerializeFailed),
-        }
+        };
 
         match self.storage.add(Key(id.to_string()), Value(val)) {
             Ok(()) => Ok(()),
@@ -89,12 +87,10 @@ impl VdcCollection {
 
     /// Get a credential from the store.
     pub fn get(&self, id: &str) -> Result<Credential, VdcCollectionError> {
-        let raw;
-
-        match self.storage.get(Key(id.to_string())) {
-            Ok(x) => raw = x.0,
+        let raw = match self.storage.get(Key(id.to_string())) {
+            Ok(x) => x.0,
             Err(_) => return Err(VdcCollectionError::LoadFailed),
-        }
+        };
 
         match serde_cbor::de::from_slice(&raw) {
             Ok(x) => Ok(x),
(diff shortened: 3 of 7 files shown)


src/vdc_collection.rs
```diff
@@ -0,0 +1,205 @@
...
+/// An individual credential.  These are mainly for internal use; access to them should go through `VdcCollection`,
+/// below.
+struct Credential {
+    id:      String,           // Probably a UUID, perhaps we should make it one?
+    format:  CredentialFormat,
+    ctype:   CredentialType,
+    tags:    Vec<String>,      // Arbitrary string tags.
+    payload: Vec<u8>,          // The actual credential.
+    dirty:   bool,             // Has this been modified since it was read from storage?
+}
```
cobward, line 28: When working in Rust we almost always use the defaults in Cargo for linting and formatting. Can you run `cargo fmt`? I'd also suggest setting up your text editor to run `cargo fmt` on save.

If you're not already, you should also look into using rust-analyzer, which is the main community LSP. To keep it in sync with your rust installation you should install using `rustup`:  `$ rustup component add rust-analyzer`.

src/vdc_collection.rs
```diff
@@ -0,0 +1,205 @@
...
+        }
+    }
+}
+
+/// Verifiable Digital Credential Collection
+///
+/// This is the main interface to credentials.
+pub struct VdcCollection {
+    list: Vec<Credential>,
+}
```
cobward, line 109: I would suggest this instead:
```suggestion
/// Verifiable Digital Credential Collection
///
/// This is the main interface to credentials.
pub struct VdcCollection {
    storage: Box<dyn StorageManagerInterface>,
}
```

Then you would only need the `add` (eager-write) and `all_*` (lazy-load) methods on this struct.

Potentially you may also want an `index`, but that depends on how you are going to layer this on top of `StorageManagerInterface`.

src/vdc_collection.rs
```diff
@@ -0,0 +1,205 @@
...
+            CredentialType::Other(x)        => println!("Other ({})", x),
+        }
+        println!("    tags:");
+        for t in self.tags.clone() {
+            println!("        {t}");
+        }
+        if self.payload.is_empty() {
+            println!("    payload:  missing?");
+        }
+    }
```
cobward, line 101: Instead of this you should `#[derive(Debug)]` on this struct and it will automatically generate equivalent code for displaying a debug view of the data.

src/vdc_collection.rs
```diff
@@ -0,0 +1,205 @@
...
+        false
+    }
+
+    /// Add a tag to a credential.
+    fn add_tag(&mut self, tag: &str) {
+        if !self.contains_tag(tag) {
+            self.tags.push(tag.to_string());
+            self.dirty = true;
+        }
+    }
```
cobward, line 66: This isn't needed for this implementation, so could you remove this, so we can treat Credentials as immutable and then we don't have to worry about consumers of this API not saving "dirty" Credentials.

src/vdc_collection.rs
```diff
@@ -0,0 +1,205 @@
+
+/// Supported credential formats.
+#[derive(PartialEq)]
+pub enum CredentialFormat {
+    MDoc,
+    W3cJwtVc,
+    W3cJwtVcJsonld,
```
cobward, line 7: Align with the format identifiers defined in the OID4VC specs.

```suggestion
    MsoMdoc,
    JwtVcJson,
    LdpVc,
```

cobward (changes requested):

src/vdc_collection.rs
todd-spruceid: Removed.

src/vdc_collection.rs
todd-spruceid: Fixed.

src/vdc_collection.rs
todd-spruceid: Done.  Rather than an index, I've put in `list()` for the storage manager.

src/vdc_collection.rs
todd-spruceid: Done.

src/vdc_collection.rs
todd-spruceid: Done.

todd-spruceid:
A sample program to exercise this code:

[main.rs.txt](https://github.com/user-attachments/files/16477865/main.rs.txt)

Put that in a new project along with `vdc_collection.rs`, `storage_manager.rs`, and `local_store.rs`.  It creates some (fake) credentials and manipulates them.

todd-spruceid:
I don't see how to re-run `uniffi-bindgen` for the swift target; I tried:
`cargo run --bin uniffi-bindgen generate --library target/debug/libmobile_sdk_rs.dylib --language swift --out-dir MobileSdkRs/Sources/MobileSdkRs` but the results changed far more than I expected.

Apparently I had to install `cargo-swift` and then do:
`cargo swift package -p ios -n MobileSdkRs`

src/vdc_collection.rs
```diff
@@ -0,0 +1,130 @@
...
+
+    /// Dump the contents of the credential set to the logger.
+    pub fn dump(&mut self) {
+        for key in self.all_entries() {
+            match self.get(&key.0) {
+                Some(x) => println!("{:?}", x),
+                None => {}
+            }
+        }
+    }
```
cobward, line 129: Use `tracing::info` instead of `println` for logging.

src/vdc_collection.rs
```diff
@@ -0,0 +1,130 @@
...
+
+impl VdcCollection {
+    /// Create a new credential set.
+    pub fn new(engine: Box<dyn StorageManagerInterface>) -> VdcCollection {
+        VdcCollection { storage: engine }
+    }
+
+    /// Add a credential to the set.
+    pub fn add(
+        &mut self,
```
cobward, line 62: You should be able to remove `mut` from all of these methods.

src/vdc_collection.rs
```diff
@@ -0,0 +1,130 @@
...
+pub enum CredentialType {
+    Iso18013_5_1mDl,
+    VehicleTitle,
+    Other(String), // For ease of expansion.
+}
+
+/// An individual credential.
+#[derive(Debug, Serialize, Deserialize)]
+pub struct Credential {
+    id: String, // Probably a UUID, perhaps we should make it one?
```
cobward, line 24: Use the `uuid::Uuid` type.

src/vdc_collection.rs
```diff
@@ -0,0 +1,130 @@
...
+
+        match serde_cbor::to_vec(&Credential::new(id, format, ctype, payload)) {
+            Ok(x) => val = x,
+            Err(_) => return,
+        }
+
+        match self.storage.add(Key(id.to_string()), Value(val)) {
+            Ok(()) => {}
+            Err(x) => println!("error adding credential: {x}"),
+        }
```
cobward, line 78: All of these methods should return [`Result`](https://doc.rust-lang.org/std/result/enum.Result.html) instead of ignoring/logging errors.

src/vdc_collection.rs
cobward, line 1: You need to add this and `local_store` as modules in [lib.rs](https://github.com/spruceid/mobile-sdk-rs/blob/main/src/lib.rs), otherwise they aren't part of the Rust library.

cobward (changes requested):

src/vdc_collection.rs
todd-spruceid: Fixed.

src/vdc_collection.rs
todd-spruceid: Done.

src/vdc_collection.rs
todd-spruceid: Fixed.

Cargo.toml
```diff
@@ -23,9 +23,12 @@ isomdl = { git = "https://github.com/spruceid/isomdl", rev = "1f4f762" }
 thiserror = "1.0.56"
 hex = "0.4.3"
 p256 = { version = "0.13.2", features = ["pkcs8"] }
+serde = "1.0.204"
 serde_cbor = "0.11.2"
 serde_json = "1.0.111"
+serde_derive = "1.0.204"
```
cobward, line 29: To reduce dependency complexity, use the `derive` feature from `serde` instead of depending on `serde_derive`.
```suggestion
serde = { version = "1.0.204", features = ["derive"] }
serde_cbor = "0.11.2"
serde_json = "1.0.111"
```

src/storage_manager.rs
```diff
@@ -56,7 +56,12 @@ pub trait StorageManagerInterface: Send + Sync {
     ///
     /// Callback function pointer to native (kotlin/swift) code for
     /// getting a key.
-    fn get(&self, key: Key) -> Result<Option<Value>, StorageManagerError>;
+    fn get(&self, key: Key) -> Result<Value, StorageManagerError>;
+
+    /// Function: list
+    ///
+    /// Callback function pointer for listing available keys.
+    fn list(&self) -> Vec<Key>;
```
cobward, line 64:
```suggestion
    fn list(&self) -> Result<Vec<Key>, StorageManagerError>;
```

src/storage_manager.rs
```diff
@@ -56,7 +56,12 @@ pub trait StorageManagerInterface: Send + Sync {
     ///
     /// Callback function pointer to native (kotlin/swift) code for
     /// getting a key.
-    fn get(&self, key: Key) -> Result<Option<Value>, StorageManagerError>;
+    fn get(&self, key: Key) -> Result<Value, StorageManagerError>;
```
cobward, line 59: The `StorageManagerError` enum does not have a variant for "not found", and I think it is more idiomatic to separate "not found" from IO errors.
```suggestion
    fn get(&self, key: Key) -> Result<Option<Value>, StorageManagerError>;
```

src/vdc_collection.rs
```diff
@@ -0,0 +1,134 @@
+use super::storage_manager::*;
+use serde_derive::{Deserialize, Serialize};
```
cobward, line 2:
```suggestion
use serde::{Deserialize, Serialize};
```

src/vdc_collection.rs
```diff
@@ -0,0 +1,134 @@
...
+        match serde_cbor::de::from_slice(&raw) {
+            Ok(x) => Ok(x),
+            Err(_) => Err(VdcCollectionError::DeserializeFailed),
+        }
+    }
+
+    /// Get a list of all the credentials.
+    pub fn all_entries(&self) -> Vec<Key> {
+        self.storage.list()
+    }
```
cobward, line 104: Doesn't this list everything in the storage, not just "credentials"?

src/vdc_collection.rs
```diff
@@ -0,0 +1,134 @@
...
+        format: CredentialFormat,
+        ctype: CredentialType,
+        payload: Vec<u8>,
+    ) -> Result<(), VdcCollectionError> {
+        let val = match serde_cbor::to_vec(&Credential::new(id, format, ctype, payload)) {
+            Ok(x) => x,
+            Err(_) => return Err(VdcCollectionError::SerializeFailed),
+        };
+
+        match self.storage.add(Key(id.to_string()), Value(val)) {
```
cobward, line 82: I think it would be prudent to use a prefix to distinguish credentials from other files/objects that could be stored.

src/vdc_collection.rs
```diff
@@ -0,0 +1,134 @@
...
+pub struct VdcCollection {
+    storage: Box<dyn StorageManagerInterface>,
+}
+
+pub enum VdcCollectionError {
+    SerializeFailed,
+    DeserializeFailed,
+    StoreFailed,
+    LoadFailed,
+}
```
cobward, line 61: Derive `thiserror::Error` and add some descriptions for each variant (see how that's done for `StorageManagerError`)

src/vdc_collection.rs
```diff
@@ -0,0 +1,134 @@
...
+/// This is the main interface to credentials.
+pub struct VdcCollection {
+    storage: Box<dyn StorageManagerInterface>,
+}
+
+pub enum VdcCollectionError {
+    SerializeFailed,
+    DeserializeFailed,
+    StoreFailed,
+    LoadFailed,
```
cobward, line 60: You can capture the previous error for better context:
```suggestion
    StoreFailed(StorageManagerError),
    LoadFailed(StorageManagerError),
```

src/vdc_collection.rs
```diff
@@ -0,0 +1,134 @@
...
+        }
+    }
+}
+
+/// Verifiable Digital Credential Collection
+///
+/// This is the main interface to credentials.
+pub struct VdcCollection {
+    storage: Box<dyn StorageManagerInterface>,
+}
```
cobward, line 54: Could you write some unit tests for the VdcCollection methods using the local store?

cobward (changes requested):
Making good progress!

src/storage_manager.rs
todd-spruceid: The possible `StorageManagerError` values are:

- invalid key
- unable to decrypt value
- storage full
- couldn't make key

None of these are applicable to getting a list of keys, so it seemed to me that either we needed to extend the error list to cover the only failure case I can think of (the underlying storage manager finding a key/value pair but being somehow unable to read the key, I guess this would be `StorageManagerErrror::SystemError` or somesuch?), or it made more sense to just return a list regardless.

From an API point of view, if the caller is asking for a list, they almost certainly just want to know what is available that can be used, so returning a list of keys seems more robust to me; otherwise we'd be returning `Err(something)` instead of the list if any of the keys were a problem, or returning a `Vec<Result<Key, Error>>` which seems like it would be a hassle to use.

cobward: In the swift implementation there seems to be an error being ignored here:
https://github.com/spruceid/mobile-sdk-swift/pull/24/files#diff-c794a03170cd0cf34fb4fa8c6f7c69791900504d65c0c769534c579a52be1106R87-R91

If it doesn't fit any of the existing errors, then a new error variant should be added.

cobward: I agree that `Vec<Result<Key, Error>>` is not helpful.

todd-spruceid: Indeed, but that is likely to be "directory doesn't exist", which would be the result if no keys had been stored yet (the directory is created lazily). In the "no keys created yet" case, at least, I think we want to return an empty list rather than a storage manager error.

I believe `contentsOfDirectory()` is wrapping `opendir()`, so in theory it can return:

- ENOENT -- isn't there, we should return an empty list
- EMFILE, ENFILE, ENOMEM -- OS resource failure, if we hit this state everything's likely gone pear-shaped and we almost certainly have larger problems
- ENOTDIR -- something took our directory name and created a regular file (or a fifo or something?) instead

The "it's not a directory" case is a corner case; something else would have had to create a regular file (or fifo or something) with the directory path we were expecting to use.  On one hand, that seems like an attempted attack; maybe they've symlinked it to some important file and are hoping we nuke it or do something that lets them read it. On the other hand it could be bad luck or misconfiguration.  Or filesystem corruption, I suppose.

For that case I'd be inclined to fix it inside the storage manager and complain to the logger rather than report an error; all the caller could do with an error is throw up a dialog informing the user that things are broken.

src/storage_manager.rs
todd-spruceid: From an API point of view, this does mean the caller has to do two error checks rather than one, and from the caller point of view I don't think the two separate error types are useful.  Personally my inclination would be to add a `StorageManagerError::NotFound` rather than make the caller check (a) if it worked, and then (b) if they actually got anything.

Cargo.toml
todd-spruceid: Fixed, will commit shortly.

src/vdc_collection.rs
todd-spruceid: Fixed, will commit shortly.

src/vdc_collection.rs
todd-spruceid: It does; are we expecting to use storage for anything else?  If so, we'll presumably need to have a way to tell stored credentials from other things; either a naming convention for keys, a subdirectory for credentials, or something stored in the value we can use to filter.

src/vdc_collection.rs
todd-spruceid: I'll do that.

src/vdc_collection.rs
todd-spruceid: Going with the prefix suggestion below.

todd-spruceid: Fixed.

src/vdc_collection.rs
todd-spruceid: Done.

src/vdc_collection.rs
todd-spruceid: Done.

src/vdc_collection.rs
todd-spruceid: Done.

src/vdc_collection.rs
todd-spruceid: Done.

src/vdc_collection.rs
```diff
@@ -0,0 +1,272 @@
...
+
+        match vdc.add(
+            uuid!("00000000-0000-0000-0000-000000000001"),
+            CredentialFormat::MsoMdoc,
+            CredentialType::Iso18013_5_1mDl,
+            payload_1.clone(),
+        ) {
+            Ok(_) => {}
+            Err(_) => panic!("Unable to add the first value."),
+        }
```
Ryanmtate, line 208: We can use `.expect` on a `Result` type to core dump with a message.

```rust
vdc.add(
     uuid!("00000000-0000-0000-0000-000000000001"),
     CredentialFormat::MsoMdoc,
     CredentialType::Iso18013_5_1mDl,
     payload_1.clone(),
).expect("Unable to add the first value.");
```

todd-spruceid: Fixed.

src/vdc_collection.rs
```diff
@@ -0,0 +1,272 @@
...
+
+            if let Ok(x) = cred {
+                if x.ctype == ctype {
+                    r.push(key.0);
+                }
+            }
+        }
+
+        r
+    }
```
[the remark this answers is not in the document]
Ryanmtate: Edit: I noticed that we're returning `key.0`, so the above example would not work.

The following example returns the inner Key String. One question maybe worth asking is whether we should return the `Key` type and have a `.deref()` method for the inner String.

```rust
self.all_entries()
            .iter()
            .filter_map(|key| {
                if let Ok(cred) = self.get(&key.0) {
                    if cred.ctype == ctype {
                        return Some(key.0.clone());
                    }
                }

                None
            })
            .collect()
```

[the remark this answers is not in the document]
todd-spruceid: I'm not entirely convinced the `Key` type is doing anything for us, given it's just a wrapper around `String`. It was in the original interface, so I left it as it was, but I don't see any benefit to keeping it as opposed to simply using `String` for keys.

src/storage_manager.rs
cobward: I'm not going to die on this hill but my opinion is that an application probably wants to handle these cases differently:
IOError -> probably not recoverable, log the specific error and move on.
NotFound -> probably recoverable, may involve user interaction depending on the context.

Is it really unusual to have a method that can throw and return a nullable value?
> func get(key: Key) throws  -> Value?

src/storage_manager.rs
cobward: The problem is if we don't bubble up the error then we will get unexpected behaviour in the frontend.

For example, this code will be used by the OpenID4VP protocol to display to the user which credentials they can present in response to a request. If we just return an empty array when there is an error then we can no longer do any handling on the frontend and in the protocol for that error.

In general I think it is responsible for libraries to bubble up any and all errors, and leave it up to the application to decide if and how it should handle the error.

src/vdc_collection.rs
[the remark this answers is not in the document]
cobward: It's just syntactic sugar for a type alias when compiled into Swift and Kotlin. Just makes the API a bit easier to use, instead of having to think about "what string am I supposed to use"?

> public typealias Key = String

src/vdc_collection.rs
```diff
@@ -0,0 +1,245 @@
...
+    /// Get a list of all the credentials.
+    pub fn all_entries(&self) -> Vec<Key> {
+        let mut r = Vec::new();
+
+        for key in self.storage.list() {
+            let name = key.0;
+
+            if name.starts_with(KEY_PREFIX) {
+                if let Some(id) = name.strip_prefix(KEY_PREFIX) {
+                    r.push(Key(id.to_string()));
```
cobward, line 138: Just return a string here, or add a [newtype](https://doc.rust-lang.org/rust-by-example/generics/new_types.html) for "credential id" here. `Key` from storage manager should be considered the path to the file. A consumer of this API should just care about the id of the credential.
```suggestion
                    r.push(id.to_string());
```

todd-spruceid: Ok, I've made these changes.

cobward (approved):

PR #25, opened 2024-08-05
w3c-vc-barcodes integration

w4ll3:
Added PDF417 verification to complement.
@Juliano1612 added MRZ + QRCode verification.


src/lib.rs
```diff
@@ -46,6 +67,200 @@ impl UniffiCustomTypeConverter for Uuid {
     }
 }
 
+pub struct StatusLists;
```
sbihel, line 70: Could you put the status list related code, and the vc-barcode code, in their own files? I thought it wasn't possible with uniffi but it actually is if you use `pub mod`

src/lib.rs
```diff
@@ -46,6 +67,200 @@ impl UniffiCustomTypeConverter for Uuid {
...
+
+impl TypedStatusMapProvider<Uri, BitstringStatusListCredential> for StatusLists {
+    async fn get_typed(&self, id: &Uri) -> Result<MaybeCached<StatusList>, ProviderError> {
```
sbihel, line 73: I'm surprised `Uri` would work as you haven't defined a wrapper type. Is this code actually used?

src/lib.rs
```diff
@@ -46,6 +67,200 @@ impl UniffiCustomTypeConverter for Uuid {
...
+        eprintln!("fetch <{id}>");
```
sbihel, line 74: Remove this print. We haven't fully figured out the logging, but we can't just use prints

src/lib.rs
```diff
@@ -46,6 +67,200 @@ impl UniffiCustomTypeConverter for Uuid {
...
+        Ok(MaybeCached::NotCached(StatusList::from_bytes(
+            1.try_into().unwrap(),
```
sbihel, line 76: let's return an error if this conversion fails

src/lib.rs
```diff
@@ -46,6 +67,200 @@ impl UniffiCustomTypeConverter for Uuid {
...
+        .map_err(|e| VCBVerificationError::Generic {
+            value: e.to_string(),
+        })?
+        .map_err(|e| VCBVerificationError::Generic {
+            value: e.to_string(),
+        })
+}
+
+fn convert_to_mrz_entry(s: &[u8]) -> &[u8; 30] {
+    s.try_into().expect("slice with incorrect length")
```
sbihel, line 133: Let's return an error instead of panicking here as well

src/lib.rs
```diff
@@ -46,6 +67,200 @@ impl UniffiCustomTypeConverter for Uuid {
...
+        AnyDidMethod::default().into_vm_resolver(),
+        status_list_client,
+    );
+
+    verify(&vc, &dl.mandatory, params)
+        .await
+        .map_err(|e| VCBVerificationError::Generic {
+            value: e.to_string(),
+        })?
+        .map_err(|e| VCBVerificationError::Generic {
```
sbihel, line 127: Could you add a new variant to say "verification failed" -- so that the app can distinguish between something going wrong and the verification simply failing

src/lib.rs
```diff
@@ -46,6 +67,200 @@ impl UniffiCustomTypeConverter for Uuid {
...
+        })?;
+
+    // Finally we verify the VCB against the MRZ data.
+    let params = VerificationParameters::new(AnyDidMethod::default().into_vm_resolver());
+    verify(&vc, &mrz, params)
+        .await
+        .map_err(|e| VCBVerificationError::Generic {
+            value: e.to_string(),
+        })?
+        .map_err(|e| VCBVerificationError::Generic {
```
sbihel, line 177: Could you add a new variant to say "verification failed" -- so that the app can distinguish between something going wrong and the verification simply failing

src/lib.rs
```diff
@@ -46,6 +67,200 @@ impl UniffiCustomTypeConverter for Uuid {
...
+    key.key_id = Some(did.into());
+
+    let jwt = vp
+        .to_jwt_claims()
+        .map_err(|e| VPError::Parsing {
+            value: e.to_string(),
+        })?
+        .sign(&key)
+        .await
+        .map_err(|e| VPError::Parsing {
```
sbihel, line 236: This needs a new variant to say that it's the signing that failed, not the parsing

src/lib.rs
```diff
@@ -46,6 +67,200 @@ impl UniffiCustomTypeConverter for Uuid {
...
+    Parsing { value: String },
+}
+
+#[uniffi::export]
+pub async fn vc_to_signed_vp(vc: String, key_str: String) -> Result<String, VPError> {
+    use ssi::prelude::*;
+
+    let vp = ssi::claims::vc::v1::JsonPresentation::new(None, None, vec![vc]);
+
+    let mut key: ssi::jwk::JWK = serde_json::from_str(&key_str).map_err(|e| VPError::Parsing {
```
sbihel, line 223: If you want, you can define a new type like with do for UUID with `uniffi::custom_type!`, but it's not necessary for now

src/lib.rs
```diff
@@ -46,6 +67,200 @@ impl UniffiCustomTypeConverter for Uuid {
...
+        value: e.to_string(),
+    })?;
+
+    let vm_resolver: ssi::dids::VerificationMethodDIDResolver<AnyDidMethod, AnyMethod> =
+        AnyDidMethod::default().into_vm_resolver();
+    let params = VerificationParameters::from_resolver(vm_resolver);
+
+    jwt.verify(params)
+        .await
+        .map_err(|e| VPError::Verification {
```
sbihel, line 256: I think this would need to be a generic error to say that something went wrong somewhere and it's not actionable?

src/lib.rs
```diff
@@ -275,6 +490,41 @@ mod tests {
 
     use super::*;
 
+    #[tokio::test]
+    async fn verify_vc() {
+        let json_vc = include_str!("../tests/res/vc");
+        let result = verify_json_vc_string(json_vc.into()).await.is_ok();
```
sbihel, line 496: In general you don't assert that a `Result` is `Ok` -- you simply `unwrap()` it which has to benefit of printing the error for debugging, but it's not a necessary change if it works

README.md
```diff
@@ -44,7 +44,7 @@ cd kotlin
 ### Swift
 
 ```bash
-cargo swift package -p ios -n MobileSdkRs
+cargo swift package -p ios -n MobileSdkRs --release
```
sbihel, line 47: May I ask if using a release build is still needed with the latest version of uniffi/cargo-swift/etc?

sbihel (changes requested):

Cargo.toml
```diff
@@ -27,11 +27,13 @@ serde = { version = "1.0.204", features = ["derive"] }
 serde_cbor = "0.11.2"
 serde_json = "1.0.111"
 uuid = { version = "1.6.1", features = ["v4"] }
+w3c-vc-barcodes = { git = "https://github.com/spruceid/w3c-vc-barcodes", rev = "d3b3b9d" }
+ssi = { git = "https://github.com/spruceid/ssi.git", rev = "88110a9", features = ["secp256r1", "secp384r1"] }
```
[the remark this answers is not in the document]
Juliano1612: I tried to use the published ssi versions, but it has conflicts with the [version used by w3c-vc-barcodes](https://github.com/spruceid/w3c-vc-barcodes/blob/main/Cargo.toml#L13)

src/lib.rs
Juliano1612: I prefer to keep it this way because we are just using it internally. If we export/return this at some point we can create a custom type

README.md
Juliano1612: I had some runtime errors while using the w3c-vc-barcodes through uniffi, and they disappeared when I compiled with the release flag

src/lib.rs
Juliano1612: `w3c_vc_barcodes::verify` requires this implementation, but since we don't have a test w3c-vc-barcode with a valid status list URI we are bypassing the retrieval with this code. 

Adding a TODO for the future.

Cargo.toml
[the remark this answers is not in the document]
sbihel: I've changed `w3c-vc-barcodes` to use the published version of `ssi`.
```toml
w3c-vc-barcodes = { git = "https://github.com/spruceid/w3c-vc-barcodes", rev = "372ff39" }
ssi = { version = "0.8.1", features = ["secp256r1", "secp384r1"] }
```

src/lib.rs
sbihel: I think I completely missed the fact that it was an trait implementation 🤦

sbihel (approved):

README.md
sbihel: `cargo-swift` recently had a patch verison released, and we're not using the latest `uniffi` version. Do you think you would have time to test again with both of these at their latest version? Otherwise, put a comment in the README to explain why we need `--release`

Juliano1612: Tested with `cargo-swift@0.8.1` and `uniffi@0.28.1`, the error is still there. I'll commit bumping the uniffi version and adding a note to the readme explaining why we are using the `--release` flag

PR #28, opened 2024-08-19
Feat/integrate oid4vp into mobile sdk rust swift kotlin

Ryanmtate:
Merge #35 first.


Ryanmtate:
NOTE: Need to fix rebased commit history. Would be nice if we can squash commits for simplifying rebasing.

NOTE: `Cargo.lock` is added to his branch. Should we add it to `.gitignore`? I noticed it had not beed added, possibly for a reason.

src/oid4vp/presentation_exchange.rs
```diff
@@ -0,0 +1,50 @@
...
+#[derive(uniffi::Object)]
+pub struct AuthorizationRequestObject(pub(crate) authorization_request::AuthorizationRequestObject);
+
+/// The [AuthorizationRequest] struct represents an authorization request
+/// sent by a verifier to a holder.
+#[derive(uniffi::Object)]
+pub struct AuthorizationRequest(authorization_request::AuthorizationRequest);
+
+#[uniffi::export]
+impl AuthorizationRequest {
```
Ryanmtate, line 17: Consider removing this since we can parse from a Url string?

src/wallet.rs
```diff
@@ -0,0 +1,127 @@
...
+
+        // Resolve the presentation definition.
+        let presentation_definition = request_object
+            .resolve_presentation_definition(self.http_client())
+            .await
+            .map_err(|e| WalletError::OID4VPPresentationDefinitionResolution(e.to_string()))?;
+
+        Ok(None)
+    }
+}
```
Ryanmtate, line 98: Processing of the presentation definition and construction of the presentation submission.

src/oid4vp/presentation_exchange.rs
```diff
@@ -0,0 +1,50 @@
...
+    #[uniffi::constructor]
+    pub fn from_query_params(params: &str) -> Result<Arc<Self>, OID4VPError> {
+        let request = authorization_request::AuthorizationRequest::from_query_params(params)
+            .map_err(|e| {
+                OID4VPError::UnexpectedUniFFICallbackError(format!(
+                    "Failed to parse authorization: {e}"
+                ))
+            })?;
+        Ok(Arc::new(Self(request)))
+    }
```
Ryanmtate, line 40: As per note above, are these methods needed to be exposed via FFI?

Currently, wallet accepts a `Url` type and can internally using the native Rust `AuthorizationRequest` methods.

Cargo.lock
Ryanmtate, line 1: Should this be added to `.gitignore`?

src/wallet.rs
```diff
@@ -0,0 +1,262 @@
...
+            .collect::<Vec<DescriptorMap>>();
+
+        // Create a presentation submission.
+        let presentation_submission = PresentationSubmission::new(
+            uuid::Uuid::new_v4(),
+            presentation_definition.parsed().id().clone(),
+            descriptor_map,
+        );
+
+        // Create a VP Token response.
```
Ryanmtate, line 189: TODO

src/wallet.rs
```diff
@@ -0,0 +1,262 @@
...
+            dyn ::core::future::Future<Output = anyhow::Result<(), anyhow::Error>>
+                + ::core::marker::Send
+                + 'async_trait,
+        >,
+    >
+    where
+        'life0: 'async_trait,
+        'life1: 'async_trait,
+        Self: ::core::marker::Sync + 'async_trait,
+    {
```
Ryanmtate, line 211: TODO: Try using `#[async_trait::async_trait]` here to clean up the lifetimes.

src/wallet.rs
```diff
@@ -0,0 +1,424 @@
...
+
+        let ldp_options = self.new_linked_data_proof_options(key_index)?;
+
+        // Constuct unsigned JWT, then sign it with the key manager.
+        let unsigned_jwt_vp = vp
+            .generate_jwt(None, &ldp_options, did_jwk::DIDJWK.to_resolver())
+            .await
+            .map_err(|e| WalletError::GenerateJwt(e.to_string()))?;
+
+        unimplemented!()
```
Ryanmtate, line 349: NOTE: working through the latest ssi-* crate API changes for generating unsigned jwt_vp. Taking inspiration from companion test wallet, but using the latest ssi-* crates. There's a few API changes in the latest versions, versus the flow used in the companion wallet based on the prior releases.

src/oid4vp/presentation_exchange.rs
Ryanmtate: Removed as extraneous.

src/wallet.rs
Ryanmtate: unencoded authorization response with the encoded JWT VP token response is now handled.

src/wallet.rs
Ryanmtate: This has been resolved.

src/wallet.rs
Ryanmtate: After fighting a bit with the versioning, I decided to use the latest `JWTClaims` API for signing.

Working on testing. This is using the latest ssi-* versions and verifiable presentation creation flow based on the oid4vp e2e test.

src/wallet.rs
Ryanmtate: This now completes the flow and handles the correct `ResponseMode`.

src/oid4vp/presentation_exchange.rs
Ryanmtate: These methods have been removed.

src/mdl/mod.rs
Ryanmtate, line 1: NOTE: This file moved the contents of `lib.rs` into a separate `mdl` module.

src/wallet.rs
```diff
@@ -0,0 +1,320 @@
...
+        let metadata = MetadataManager::initialize(&storage_manager)?;
+
+        Ok(Arc::new(Self {
+            client,
+            metadata,
+            key_manager,
+            storage_manager,
+            vdc_collection: VdcCollection::new(),
+            trust_manager: TrustManager::new(),
+            active_key_index: Arc::new(RwLock::new(DEFAULT_KEY_INDEX)),
```
Ryanmtate, line 163: Consider creating a `KeyRing` struct to maintain the active key index and ensure that the device has at least one signing key available. Creating a default key if one does not exist.

kotlin/mobilesdkrs/build.gradle
```diff
@@ -161,10 +163,10 @@ signing {
...
-        publish("release") {
-            it.username.set(System.getenv("MAVEN_USERNAME"))
-            it.password.set(System.getenv("MAVEN_PASSWORD"))
-            it.publicationType.set("AUTOMATIC")
-        }
+        // publish("release") {
+        //     it.username.set(System.getenv("MAVEN_USERNAME"))
+        //     it.password.set(System.getenv("MAVEN_PASSWORD"))
+        //     it.publicationType.set("AUTOMATIC")
+        // }
```
Ryanmtate, line 170: TODO: uncomment out.

README.md
```diff
@@ -32,24 +32,17 @@ security firm before the v1.0 release.
...
-```bash
-cargo swift package -p ios -n MobileSdkRs
-```
 > You will need `cargo-swift` which you can install with `cargo install cargo-swift`.
 
-## Test
-In order to run the tests you'll need to download a copy of JNA
+-  **Download and link `JNA`**
+
+> In order to run the tests you'll need to download a copy of JNA
```
Ryanmtate, line 45: TODO: Merge latest README changes, e.g., kotlinx coroutines setup.

kotlin/build.gradle
```diff
@@ -1,7 +1,7 @@
 plugins {
     id 'com.android.application' version '8.1.2' apply false
     id 'com.android.library' version '8.1.2' apply false
-    id 'org.jetbrains.kotlin.android' version '1.9.10' apply false
+    id 'org.jetbrains.kotlin.android' version '2.0.0' apply false
```
Ryanmtate, line 4: TODO: Revert changes ? (need to ensure there is nothing needed in bumping this version).

@todd-spruceid brought up a good point that we should try to maintain legacy versions (when we don't need API updates, etc.) so that we can cover the broadest (older) phone versions in the population of users.

kotlin/mobilesdkrs/build.gradle
```diff
@@ -46,8 +46,10 @@ dependencies {
     androidTestImplementation 'androidx.test.ext:junit:1.1.5'
     androidTestImplementation 'androidx.test.espresso:espresso-core:3.5.1'
 
-    implementation "net.java.dev.jna:jna:5.13.0@aar"
-    implementation "org.jetbrains.kotlinx:kotlinx-coroutines-core:1.6.4"
+    implementation 'net.java.dev.jna:jna:5.13.0@aar'
+    implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-core:1.9.0-RC.2'
+    implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-android:1.9.0-RC.2'
```
Ryanmtate, line 51: TODO: Revert changes to `1.6.4`. See previous comments for reasoning. (i.e. support more phones/users)

kotlin/mobilesdkrs/build.gradle
```diff
@@ -83,7 +85,7 @@ tasks.register('bindGen', Exec) {
     workingDir "../../"
     if (Os.isFamily(Os.FAMILY_WINDOWS)) {
         commandLine("cmd", "/c",
-                "cargo build && " + "cargo run --bin uniffi-bindgen generate --library target\\debug\\libmobile_sdk_rs.dll " + "--language kotlin " + "--out-dir " + outDir.replace('/', '\\'))
+                "cargo build --release && " + "cargo run --bin uniffi-bindgen generate --library target\\release\\libmobile_sdk_rs.dll " + "--language kotlin " + "--out-dir " + outDir.replace('/', '\\'))
```
Ryanmtate, line 88: Any reason to use `debug` instead of `release` here? Safe to update?

kotlin/mobilesdkrs/src/androidTest/java/com/spruceid/mobile/sdk/rs/ExampleInstrumentedTest.kt
Ryanmtate, line 1: TODO: Given the classpath solution to using kotlinx in uniffi bindgen tests, move this file to the test harness.

src/key_manager.rs
```diff
@@ -0,0 +1,113 @@
...
+        KeyManagerError::UnexpectedUniFFICallbackError(value.reason)
+    }
+}
+
+/// EncryptedPayload is a struct that holds the IV and ciphertext
+/// of an encrypted payload.
+#[derive(uniffi::Object, Debug, Clone)]
+pub struct EncryptedPayload {
+    iv: Vec<u8>,
+    ciphertext: Vec<u8>,
```
Ryanmtate, line 52: TODO: Do we want to expand this type to support key exchange DH encryption? Would include the public key of the intended recipient as a field, possibly optional.

We would also maybe consider expanding the `KeyManager` interface to include encrypting using a recipient's public key. This would be useful for encryption in-transit purposes, versus local encryption at-rest only.

src/key_manager.rs
```diff
@@ -0,0 +1,113 @@
...
+    fn generate_signing_key(&self, id: Key) -> bool;
+
+    /// Return a JWK for a given key ID as a JSON-encoded string.
+    fn get_jwk(&self, id: Key) -> Result<String, KeyManagerError>;
+
+    /// Sign a payload with a key in the key manager.
+    fn sign_payload(&self, id: Key, payload: Vec<u8>) -> Result<Vec<u8>, KeyManagerError>;
+
+    /// Generate an encryption key in the key manager.
+    fn generate_encryption_key(&self, id: Key) -> bool;
```
Ryanmtate, line 97: This may already support DH key exchange, if implemented correctly. We'd then just need to ensure the Encryption Payload has the public key to match to the corresponding private key for decryption.

src/wallet.rs
```diff
@@ -0,0 +1,219 @@
...
+/// [Verifiable Credentials](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#section-2-3.4) and keys can
+/// both be stored/managed locally, or by using a remote self-hosted service, or a remote third-party service.
+///
+/// In the context of the [OID4VP](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html) specification,
+/// the Wallet acts as an OAuth 2.0 Authorization Server
+/// (see [RFC6749](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#RFC6749))
+/// towards the Credential Verifier which acts as the OAuth 2.0 Client.
+///
+#[derive(uniffi::Object, Debug)]
+pub struct Wallet {
```
Ryanmtate, line 98: TODO: Rename `Wallet` to `SpruceKit` or `SpruceSDK` to keep it generic for top-level mobile-sdk-rs driver. This falls in alignment with the renaming of the `wallet-sdk` -> `mobile-sdk`.

cobward: FYI wallet-sdk was renamed to mobile-sdk because we want the SDK to encapsulate more than just wallet capabilities. So I don't think that's enough of a reason to rename it.

src/trust_manager.rs
```diff
@@ -0,0 +1,288 @@
...
+/// TrustManager is responsible for managing trusted DIDs for the wallet.
+///
+/// Use the [TrustManager::new] method to create a new instance of the trust manager.
+///
+/// The trust manager does not store a cached state of the trusted dids,
+/// but instead accesses and modifies the trusted dids in the storage manager directly.
+///
+/// In the future, this might change in favor of faster reads.
+#[derive(Debug)]
+pub struct TrustManager;
```
cobward, line 33: This really isn't a bad idea, but there's a huge **_but_**. We can't just add a DID to this manager and assume every transaction we do with that DID can be trusted. For example, we might trust `did:web:suspisciousonlineshop.com` for verifying a proof-of-age credential, but we probably don't trust it with PII heavy credentials.

We might want to build something in this shape in the future, but it would need an extensive "capabilities" interface that allows a wallet to define what kinds of transactions a DID is trusted for. Whether or not a wallet trusts a verifier has a lot to do with the context of the transaction. For those reasons, I think this kind of interface is too much work and should be out of scope for this task. Instead, we should just expose an interface for an implementor to decide whether or not to trust a specific DID at transaction time.

Makefile
cobward, line 1: We usually rely on documenting the ways to build our software using the README. If you think there's enough complexity that we need a Makefile, could you use [cargo-make](https://github.com/sagiegurari/cargo-make) instead? That way, we stay within the Rust ecosystem and, as an added bonus, get a build recipe that is more obvious to readers who are not familiar with the nuances of Make.

kotlin/build.gradle
cobward: Does bumping this plugin version actually impact end users? My understanding is that this is just the Kotlin version, which shouldn't impact the compiled application.

kotlin/mobilesdkrs/build.gradle
cobward: I'm not a gradle expert. If this is only used for release builds then yes, but we don't want to do optimisations on debug builds.

src/key_manager.rs
```diff
@@ -0,0 +1,113 @@
...
+
+    // /// Encrypt a payload with a key in the key manager.
```
cobward, line 99:
```suggestion
    /// Encrypt a payload with a key in the key manager.
```

src/key_manager.rs
cobward: In what context is this method used at the moment?

src/metadata_manager.rs
```diff
@@ -0,0 +1,133 @@
...
+    /// This is an alias for [MetadataManager::as_ref].
+    pub fn cache(&self) -> &WalletMetadata {
+        self.as_ref()
+    }
+
+    // Internal method for adding wallet metadata to the wallet.
+    fn add_metadata(
+        metadata_value: &WalletMetadata,
+        storage: &Box<dyn StorageManagerInterface>,
+    ) -> Result<(), MetadataManagerError> {
```
cobward, line 110: If this method fails then the cache needs to be invalidated.

src/mdl/mod.rs
Ryanmtate: merge #30 before this.

src/key_manager.rs
Ryanmtate: This is likely premature, we can revisit when those needs are required. I'm not sure if there is a handshake in other protocols that may need such an interface, but when the need arises, we can add it. Resolving for now.

src/trust_manager.rs
Ryanmtate: Tracking this conversation in #33

src/metadata_manager.rs
Ryanmtate: Tracking this in #34

src/key_manager.rs
Ryanmtate: Not needed here.

src/credentials_callback.rs
```diff
@@ -0,0 +1,41 @@
...
+    /// The native client should implement this method, returning a vector of booleans indicating
+    /// whether the requested information can be presented.
+    ///
+    /// the format of the requested_information is a vector of strings, each string is a field name
+    /// that the verifier is requesting, e.g. ["Name", "Date Of Birth", "Address"]
+    ///
+    /// If the user denies the a field request, the verifier should return a [CredentialCallbackError::PermissionDenied]
+    fn permit_presentation(
+        &self,
+        requested_fields: Vec<String>,
```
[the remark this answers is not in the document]
Ryanmtate: `RequestedField` type has been added to provide presentation definition constraints.

Open future PR to provide a shared interface with the mdl presentation, i.e. `ItemsRequested`.

src/wallet.rs
Ryanmtate: Save for a later PR.

Makefile
Ryanmtate: I'll add a cargo-make equivalent of the Makefile in a future PR to not block upstream PRs based on this branch.

src/wallet.rs
Ryanmtate: Not needed here.

Cargo.lock
Ryanmtate: I'll leave the lock file committed in this PR, unless other's disagree.

kotlin/mobilesdkrs/src/androidTest/java/com/spruceid/mobile/sdk/rs/ExampleInstrumentedTest.kt
Ryanmtate: Migrate in a subsequent PR.

Ryanmtate:
Cherry pick oid4vp changes, refactor for incoming VDC collection changes.

Rebase on oid4vp-rs once merged.

PR #29, opened 2024-08-22
OID4VCI Integration

theosirian:
Depends on https://github.com/spruceid/oid4vci-rs/pull/18.

Bindings compilation status:
- [x] iOS
- [x] Android

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,8 +22,6 @@ isomdl = { git = "https://github.com/spruceid/isomdl", rev = "1f4f762" }
 oid4vci = { git = "https://github.com/spruceid/oid4vci-rs", branch = "feat/updates-for-sprucekit-integration" }
 ssi = { version = "0.8.1", features = ["secp256r1", "secp384r1"] }
 
-ssi = "0.8.1"
-
 async-trait = "0.1"
 either = "1.13"
 futures = "0.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,11 +22,10 @@ isomdl = { git = "https://github.com/spruceid/isomdl", rev = "1f4f762" }
 oid4vci = { git = "https://github.com/spruceid/oid4vci-rs", branch = "feat/updates-for-sprucekit-integration" }
 ssi = { version = "0.8.1", features = ["secp256r1", "secp384r1"] }
 
-ssi-claims = "0.1.0"
-ssi-dids = "0.2.0"
-ssi-dids-core = "0.1.0"
-ssi-jwk = "0.2.1"
+ssi = "0.8.1"
 
+async-trait = "0.1"
+either = "1.13"
 futures = "0.3"
 hex = "0.4.3"
 p256 = { version = "0.13.2", features = ["pkcs8"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ path = "uniffi-bindgen.rs"
 uniffi = { version = "0.28.1", features = ["cli"] }
 
 isomdl = { git = "https://github.com/spruceid/isomdl", rev = "1f4f762" }
-oid4vci = { git = "https://github.com/spruceid/oid4vci-rs", branch = "feat/updates-for-sprucekit-integration" }
+oid4vci = { git = "https://github.com/spruceid/oid4vci-rs", rev = "020887f" }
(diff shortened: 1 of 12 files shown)


Cargo.toml
```diff
@@ -16,23 +16,32 @@ name = "uniffi-bindgen"
...
 
 [dependencies]
-uniffi = { version = "0.27.1", features = ["cli"] }
+uniffi = { version = "0.28.1", features = ["cli"] }
 
 isomdl = { git = "https://github.com/spruceid/isomdl", rev = "1f4f762" }
+oid4vci = { git = "https://github.com/spruceid/oid4vci-rs", branch = "feat/updates-for-sprucekit-integration" }
 
-thiserror = "1.0.56"
+ssi-claims = "0.1.0"
```
sbihel, line 24: You can import the whole of `ssi` in this project

src/oid4vci/http_client.rs
```diff
@@ -0,0 +1,151 @@
...
+
+        response
+            .body(self.body)
+            .map_err(|_| HttpClientError::ResponseBuilder)
+    }
+}
+
+// TODO figure out how to use async_trait to have an async version of HttpClient
+#[uniffi::export(with_foreign)]
+pub trait HttpClient: Send + Sync {
```
sbihel, line 124: IMO let's not worry about doing HTTP calls from the native code, and just use `oid4vci::openidconnect::reqwest::Client`

sbihel (commented):
The CI is failing because of async (which I've fixed in https://github.com/spruceid/mobile-sdk-rs/pull/25)

There's more code than I was expecting but it all looks good. Maybe in the future we'll realise it doesn't need to be too flexible.

I won't approve until the CI is passing and it's all tested end-to-end

src/oid4vci/http_client.rs
theosirian: I went with a trait because it'll be useful to let the library user potentially configure different parameters for an HTTP client that are not specific to the OID4VCI spec, like extra headers, timeouts, certificates, etc.
But we can definitely have `oid4vci::openidconnect::reqwest::Client` be the default.

sbihel: Agreed. We could offer a default with `reqwest` but let's do that in a separate PR if we find the need for it. It doesn't seem hard to implement the HTTP client in Swift/Kotlin and offers better debugging capabilities.

src/oid4vci/proof_of_possession.rs
```diff
@@ -0,0 +1,71 @@
...
+        &pop_params,
+        duration_in_secs
+            .map(time::Duration::seconds)
+            .unwrap_or(time::Duration::minutes(5)),
+    )
+    .to_jwt_signing_input()
+    .map_err(|e| e.to_string())
+    .map_err(OID4VCIError::from)?;
+
+    Ok(signing_input)
```
Ryanmtate, line 54: #28 has an implementation of a `[KeyManager](https://github.com/spruceid/mobile-sdk-rs/pull/28/files#diff-741fdf60909022c1baae65c0125d16f0a2ed8b4f2bcab2a25eb5ec1ba265912b)` callback interface that may be worth exploring for use here.

src/oid4vci/mod.rs
```diff
@@ -0,0 +1,366 @@
...
+        VerificationParameters,
+    },
+    dids::{AnyDidMethod, DIDResolver},
+};
+
+use crate::vdc_collection::CredentialFormat;
+
+#[uniffi::export]
+pub async fn oid4vci_initiate_with_offer(
+    credential_offer: String,
```
cobward, line 47: What's the format of this string? The code implies that it's JSON, but a credential offer is parsed from a URI. Could we just pass the URI here instead?

cobward (commented):
I would also like to see some consideration for the VdcCollection and key management. How does a credential get stored in the VdcCollection, and how is that associated with a keypair that can be accessed by credential presentation code.

Ryanmtate:
@sbihel @cobward 

After syncing with @theosirian, we're going to try to merge this as-is before rebasing on the open draft PRs.

We'll upstream the changes for the `KeyManager` and the `common` types after this is merged.

Would it be okay to get this final review today for merge?

sbihel (approved):

PR #31, opened 2024-08-27
Feat/common

Ryanmtate:
Rebase #30 first.

The `common.rs` module contains common types that will be used by several different interfaces within the sdk. For example, it includes uniffi conversions of common types, like `Url` and `Uuid` and moves the `Key` and `Value` types from the storage manager, so they can be re-used.

As we identify more types that make sense to go into common, e.g. `ClaimFormatDesignation`, etc., this would be the logical place to migrate those types.

PR #32, opened 2024-08-27
update vdc collection to use a reference to the storage manager

Ryanmtate:
merge #31 and rebase on main before merging.


Ryanmtate:
TODO: merge in @rschulman `async-storage-manager` branch and ensure vdc collection methods are async.

.gitignore
```diff
@@ -1,3 +1,4 @@
+.DS_Store
```
wyc, line 1: thank goodness

Ryanmtate:
> TODO: merge in @rschulman `async-storage-manager` branch and ensure vdc collection methods are async.

We'll open a new PR with the async changes. There are additional changes that need to get populated through the wallet with some refactoring to be done upstream. This branch provides enough shared functionality to incorporate future changes later, versus in this PR.

src/vdc_collection.rs
```diff
@@ -1,197 +1,193 @@
...
+    pub fn all_entries(
+        &self,
+        storage: &Arc<dyn StorageManagerInterface>,
+    ) -> Result<Vec<Key>, VdcCollectionError> {
+        Ok(storage
+            .list()?
+            .iter()
+            .filter(|key| key.0.contains(KEY_PREFIX))
+            .map(ToOwned::to_owned)
+            .collect())
```
rschulman, line 162: This replacement is no longer stripping `KEY_PREFIX` from the key. Is that desired? Are there other places that are relying on the prefix being stripped?

rschulman (changes requested):
This looks broadly in good shape to me. Made a couple particular comments.

src/vdc_collection.rs
```diff
@@ -1,197 +1,193 @@
...
-    fn id_to_key(&self, id: Uuid) -> Key {
-        self.str_to_key(&id.to_string())
-    }
-
-    /// Convert a string ref to a storage key.
-    fn str_to_key(&self, id: &str) -> Key {
-        Key(format!("{}{}", KEY_PREFIX, id))
+    /// Return the storage prefix, given a credential type.
+    pub fn storage_prefix(ctype: &CredentialType) -> String {
+        format!("{KEY_PREFIX}{}.", ctype)
```
cobward, line 181: I don't think we should add metadata about the credential to the key, it makes it more difficult to discover credentials, maintain and migrate at a later date.

src/vdc_collection.rs
```diff
@@ -1,197 +1,193 @@
...
 }
 
 /// Verifiable Digital Credential Collection
 ///
 /// This is the main interface to credentials.
-pub struct VdcCollection {
-    storage: Box<dyn StorageManagerInterface>,
-}
+#[derive(Debug)]
+pub struct VdcCollection;
```
cobward, line 74: What's the motivation for this change?

rschulman: I'm not Ryan, but I think the reasoning here was to avoid dealing with the memory management of passing around the Box. Since VdcCollection is used from the Wallet, which stores its own `StorageManagerInterface`, we can pass in a reference to the storage manager when we need it instead of storing it permanently in the VdcCollection. See the changes to `add`, `get`, etc below.

src/vdc_collection.rs
rschulman: The reasoning here is to be able to ask the VdcCollection to list all credentials of a given type. If we store the type of the credential in the storage path, we can filter on that later, in `all_entries_by_type`.

cobward: There's lots of things that we might need to filter on. If we need an effective way for lookup then we should maintain an external metadata map in my opinion.

src/vdc_collection.rs
cobward: It's not clear to me how this is a problem, it just moves the reference to the storage manager from the constructor to the methods. AFAIK there's no complexity with memory management - when the VdcCollection is constructed we have a reference to the storage manager, and when it goes out of scope / is GC'd, the reference is destroyed. Is there some nuance I'm missing?

src/vdc_collection.rs
rschulman: I mean, if I'm honest I think its unlikely that any individual device is going to store so many credentials that we can't just return an entire list and let the consumer deal with filtering it, but I'm not tied to this usage and I haven't had a reason to use it yet.

@Ryanmtate was this crucial for anything in particular that you want to speak up for it?

src/vdc_collection.rs
rschulman: I'll let @Ryanmtate respond, but it also occurred to me that this might be a change related to moving away from using the callback interface from Uniffi, which required passing everything around in `Box`es, and instead being able to use regular borrows, but that storing a borrow in the struct itself would have caused an explosion of lifetimes. Again, I don't know for sure, though.

src/common.rs
```diff
@@ -12,7 +12,7 @@ impl UniffiCustomTypeConverter for CredentialType {
         Ok(CredentialType::from(credential_type.as_str()))
     }
     fn from_custom(credential_type: Self) -> Self::Builtin {
-        credential_type.into()
+        (&credential_type).into()
```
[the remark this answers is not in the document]
Ryanmtate: Yeah, I think I added `From<&CredentialType>` for String, which does not consume. But we can change to this so that it is consumed.

PR #35, opened 2024-08-27
add key manager callback interface

Ryanmtate:
merge #32 first.

### DIFF
--- a/src/key_manager.rs
+++ b/src/key_manager.rs
@@ -8,7 +8,7 @@ pub const KEY_MANAGER_PREFIX: &str = "SSIKeyManager.";
 /// This is the default index for a key in the key manager.
 pub const DEFAULT_KEY_INDEX: u8 = 0;
 
-#[uniffi::export(callback_interface)]
+#[uniffi::export(with_foreign)]
 pub trait SecretKeyInterface: Send + Sync {
     // TODO: Review available methods in android and cryptokit to implement here
 }
@@ -55,17 +55,17 @@ pub struct EncryptedPayload {
 #[uniffi::export]
 impl EncryptedPayload {
     #[uniffi::constructor]
-    pub fn new(iv: Vec<u8>, ciphertext: Vec<u8>) -> Self {
-        Self { iv, ciphertext }
+    pub fn new(iv: Vec<u8>, ciphertext: Vec<u8>) -> Arc<Self> {
+        Arc::new(Self { iv, ciphertext })
     }
 
     /// Get the initialization vector (IV) for the encrypted payload.
-    pub fn iv(&self) -> Vec<u8> {
+    pub fn iv(self: Arc<Self>) -> Vec<u8> {
         self.iv.clone()
     }
 
     /// Get the ciphertext for the encrypted payload.
-    pub fn ciphertext(&self) -> Vec<u8> {
+    pub fn ciphertext(self: Arc<Self>) -> Vec<u8> {
         self.ciphertext.clone()
     }
 }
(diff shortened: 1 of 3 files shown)


src/key_manager.rs
```diff
@@ -0,0 +1,113 @@
...
+    fn key_exists(&self, id: Key) -> bool;
+
+    /// Get a secret key from the key manager.
+    fn get_secret_key(&self, id: Key) -> Option<Arc<dyn SecretKeyInterface>>;
+
+    /// Generate a signing key in the key manager.
+    fn generate_signing_key(&self, id: Key) -> Result<bool, KeyManagerError>;
+
+    /// Return a JWK for a given key ID as a JSON-encoded string.
+    fn get_jwk(&self, id: Key) -> Result<String, KeyManagerError>;
```
cobward, line 91:
```suggestion
    /// Return a public JWK for a given key ID as a JSON-encoded string.
    fn get_public_jwk(&self, id: Key) -> Result<String, KeyManagerError>;
```

src/key_manager.rs
```diff
@@ -0,0 +1,113 @@
...
+        payload: Vec<u8>,
+    ) -> Result<Arc<EncryptedPayload>, KeyManagerError>;
+
+    /// Decrypt a ciphertext with a key in the key manager. Returns a
+    /// plaintext payload, if the ID exists and the decryption is successful.
+    fn decrypt_payload(
+        &self,
+        id: Key,
+        encrypted_payload: Arc<EncryptedPayload>,
+    ) -> Result<Vec<u8>, KeyManagerError>;
```
cobward, line 112: This interface doesn't have enough detail. What type of key is being created? What encryption algorithm is being used?

src/key_manager.rs
```diff
@@ -0,0 +1,113 @@
...
+    fn reset(&self) -> bool;
+
+    /// Check if a key exists in the key manager.
+    fn key_exists(&self, id: Key) -> bool;
+
+    /// Get a secret key from the key manager.
+    fn get_secret_key(&self, id: Key) -> Option<Arc<dyn SecretKeyInterface>>;
+
+    /// Generate a signing key in the key manager.
+    fn generate_signing_key(&self, id: Key) -> Result<bool, KeyManagerError>;
```
cobward, line 88: What kind of key is being created? P-256? ED25519?

src/key_manager.rs
```diff
@@ -0,0 +1,113 @@
...
+
+    /// Return a JWK for a given key ID as a JSON-encoded string.
+    fn get_jwk(&self, id: Key) -> Result<String, KeyManagerError>;
+
+    /// Sign a payload with a key in the key manager.
+    fn sign_payload(&self, id: Key, payload: Vec<u8>) -> Result<Vec<u8>, KeyManagerError>;
```
cobward, line 94: What signing algorithm is being used?

src/key_manager.rs
```diff
@@ -0,0 +1,113 @@
...
+#[uniffi::export(with_foreign)]
+pub trait KeyManagerInterface: Send + Sync + Debug {
+    /// Reset the key manager, removing all keys.
+    fn reset(&self) -> bool;
+
+    /// Check if a key exists in the key manager.
+    fn key_exists(&self, id: Key) -> bool;
+
+    /// Get a secret key from the key manager.
+    fn get_secret_key(&self, id: Key) -> Option<Arc<dyn SecretKeyInterface>>;
```
cobward, line 85: I think this should be removed for now, it's not clear what the purpose is. The `KeyManagerInterface` itself has methods for signing and encrypting.

cobward (changes requested):

PR #42, opened 2024-09-16
Serve MDL Holder Verification From VDC Collection

rschulman:
This PR modifies MDL Holder presentation to work from credentials held in the VDC Collection.


uniffi.toml
```diff
@@ -1,4 +1,4 @@
 [bindings.kotlin]
 package_name = "com.spruceid.mobile.sdk.rs"
 # Disabling Android optimisations due to uniffi tests breaking on missing import android
-android = false
+android = true
```
cobward, line 4: Based on that comment, I think this needs to be false.

flake.lock
cobward, line 1: Can you add the `flake.*` files to `.gitignore`?

kotlin/mobilesdkrs/build.gradle
```diff
@@ -160,11 +160,11 @@ signing {
...
-            it.publicationType.set("AUTOMATIC")
-        }
-    }
+    //afterEvaluate {
+        //publish("release") {
+            //it.username.set(System.getenv("MAVEN_USERNAME"))
+            //it.password.set(System.getenv("MAVEN_PASSWORD"))
+            //it.publicationType.set("AUTOMATIC")
+        //}
+    //}
```
cobward, line 169: This change shouldn't be necessary. If you were having trouble building locally, try out the new command in the README (`./gradlew buildCargoNdkDebug`)

src/vdc_collection.rs
```diff
@@ -160,6 +160,9 @@ mod tests {
     async fn test_vdc() {
         let smi: Arc<dyn StorageManagerInterface> = Arc::new(LocalStore);
         let vdc = VdcCollection::new(smi);
+        for id in vdc.all_entries().await.unwrap() {
+            vdc.delete(id).await.unwrap();
+        }
```
cobward, line 165: I wonder if we should use a `MemoryStore` rather than `LocalStore` for unit tests to avoid this problem entirely?

src/mdl/holder.rs
```diff
@@ -0,0 +1,491 @@
...
+        session, BleOptions, DeviceRetrievalMethod, SessionEstablishment,
+    },
+    presentation::device::{self, SessionManagerInit},
+};
+
+#[uniffi::export]
+pub async fn initialize_mdl_presentation(
+    uuid: Uuid,
+    storage_manager: Arc<dyn StorageManagerInterface>,
+) -> Result<MdlPresentationSession, SessionError> {
```
cobward, line 60: As discussed out-of-band, we want to be able to initialize using the StorageManager, or using an mdoc directly, to enable multiple migration paths.

uniffi.toml
rschulman: Ah, I forgot to remove the comment. I don't believe this is correct anymore. See https://github.com/mozilla/uniffi-rs/pull/1997 and https://github.com/mozilla/uniffi-rs/issues/1996 which I believe fixed this.

flake.lock
rschulman: Flakes really want to be in the git repo. There's ways around it, but I included them in case others want to use them as well. I can remove them if desired.

cobward: Most of the developer base won't be using nix, so I'd prefer to keep it out of the git repo.

> Flakes really want to be in the git repo.

Why is that?

rschulman: > Why is that?

I think it has something to do with pinning all of the dependencies, but I honestly don't know and find it a bit annoying. I'll remove them from git.

src/mdl/holder.rs
```diff
@@ -0,0 +1,536 @@
...
+
+    let mdoc = crate::credential::mdoc::Mdoc::from_cbor_encoded_document(
+        document.payload,
+        document.key_alias.ok_or(SessionError::Generic {
+            value: "key_alias missing in vdc collection".to_string(),
+        })?,
+    )
+    .map_err(|e| SessionError::Generic {
+        value: format!("Error retrieving MDoc from storage: {e:}"),
+    })?;
```
cobward, line 71: This isn't wrong, but the intended method for this is [`ParsedCredential::parse_from_credential`](https://github.com/spruceid/mobile-sdk-rs/blob/main/src/credential/mod.rs#L82-L92).

cobward, line 71: This isn't wrong, but the intended method for transforming a VdcCollection credential into a known type is [`ParsedCredential::parse_from_credential`](https://github.com/spruceid/mobile-sdk-rs/blob/main/src/credential/mod.rs#L82-L92), or you can just do:
```suggestion
    let mdoc: Arc<Mdoc> = document.try_into()
    .map_err(|e| SessionError::Generic {
        value: format!("Error retrieving MDoc from storage: {e:}"),
    })?;
```

src/mdl/holder.rs
```diff
@@ -0,0 +1,536 @@
...
+        // See https://doc.rust-lang.org/std/sync/struct.Mutex.html
+        let (session_manager, items_requests) = {
+            let session_establishment: SessionEstablishment = serde_cbor::from_slice(&request)
+                .map_err(|e| RequestError::Generic {
+                    value: format!("Could not deserialize request: {e:?}"),
+                })?;
+            let engaged = self
+                .engaged
+                .lock()
+                .expect("Could not lock mutex")
```
cobward, line 190: Can you return an error instead of panicking?

src/mdl/holder.rs
```diff
@@ -0,0 +1,536 @@
...
+        };
+
+        let mut in_process = self.in_process.lock().unwrap();
+        *in_process = Some(InProcessRecord {
+            session: session_manager,
+            items_request: items_requests.clone(),
+        });
+        Ok(in_process
+            .as_ref()
+            .unwrap()
```
cobward, line 208: Can you return an error instead of panicking?

cobward (changes requested):
Mostly LGTM, just a few small comments.

src/mdl/holder.rs
rschulman: Thanks, I've made that change!

src/mdl/holder.rs
rschulman: I moved a comment from elsewhere (it had been shifted around by other edits, I guess) explaining why panicking is the correct thing to do when locking a Mutex fails. Happy to change this if you disagree, however.

src/mdl/holder.rs
rschulman: This unwrap can't panic because we set `in_process` to `Some()` in the previous line. I've added a comment to that effect.

src/mdl/holder.rs
```diff
@@ -0,0 +1,533 @@
...
+        engaged: Mutex::new(Some(engaged_state)),
+        in_process: Mutex::new(None),
+        qr_code_uri,
+        ble_ident,
+    })
+}
+
+#[derive(uniffi::Object)]
+pub struct MdlPresentationSession {
+    engaged: Mutex<Option<device::SessionManagerEngaged>>,
```
cobward, line 154: Does this need to be an Option? It seems to always be constructed with `Some`

src/mdl/holder.rs
```diff
@@ -0,0 +1,533 @@
...
+                .take();
+            engaged
+                .unwrap()
+                .process_session_establishment(session_establishment)
+                .map_err(|e| RequestError::Generic {
+                    value: format!("Could not process process session establishment: {e:?}"),
+                })?
+        };
+
+        let mut in_process = self.in_process.lock().unwrap();
```
cobward, line 196: Same comment as for the above mutex.

src/mdl/holder.rs
```diff
@@ -0,0 +1,533 @@
...
+            // If that has happened, its probably better to just crash. This is what the standard documentation recommends.
+            // See https://doc.rust-lang.org/std/sync/struct.Mutex.html
+            let engaged = self
+                .engaged
+                .lock()
+                .expect("Could not lock mutex")
+                .deref_mut()
+                .take();
+            engaged
+                .unwrap()
```
cobward, line 189: If you can remove the `Option` from `self.engaged` then this is a moot point, otherwise could you add a comment of the form `// Unwrap safety: <reason>` on the line directly above the unwrap call?

src/mdl/holder.rs
rschulman: Yes, I just noticed the same thing. Left over from some old code, I've removed it.

src/mdl/holder.rs
rschulman: I've realized that I don't need to do this at all. `items_request` is still in scope at this point, so I can use it directly, avoiding the unwrap entirely.

cobward (approved):

PR #43, opened 2024-09-19
OID4VP

Ryanmtate:


src/credential/mod.rs
```diff
@@ -45,24 +45,22 @@ impl Credential {
...
+impl TryFrom<Credential> for AnyJsonCredential<NonEmptyObject> {
+    type Error = CredentialDecodingError;
+
+    fn try_from(value: Credential) -> Result<Self, Self::Error> {
+        match value.format {
+            CredentialFormat::JwtVcJson => {
+                let jwt_vc: Arc<JwtVc> = value.try_into()?;
+                Ok(jwt_vc.credential())
+            }
+            // TODO: Add more formats here.
```
Ryanmtate, line 57: Add more conversions here.

src/credential/mod.rs
```diff
@@ -44,6 +50,77 @@ impl Credential {
...
+                    .constraints()
+                    .fields()
+                    .iter()
+                    .filter_map(|field| match field.validator() {
+                        Some(Ok(validator)) => {
+                            field.path().iter().map(|path| {
+                                selector(path)
+                            })
+                            .map(|value| {
+
```
Ryanmtate, line 105: WIP: Apply constraint validator to credential

kotlin/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,7 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.0-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.5-bin.zip
```
Ryanmtate, line 3: Consider bumping this to `8.7`, which is what is being used in `mobile-sdk-kt`.

kotlin/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,7 @@
...
 networkTimeout=10000
+validateDistributionUrl=true
```
Ryanmtate, line 5: This may not be necessary?

kotlin/mobilesdkrs/build.gradle
```diff
@@ -116,8 +116,17 @@ publishing {
                 password = System.getenv("GITHUB_TOKEN")
             }
         }
+        mavenLocal()
```
Ryanmtate, line 119: NOTE: Adding `mavenLocal` repository here, which I was able to check is publishing to `~/.m2/` local repository.

TODO: need to add the following to the README for publishing to maven local:

```bash
VERSION=SEMVER_VERSION_HERE ./gradlew publishDebugPublicationToMavenLocal
```

src/credential/mod.rs
Ryanmtate: TODO: add Mdoc presentation capabilities. This will be in a follow up PR.

src/credential/mod.rs
```diff
@@ -3,27 +3,258 @@ pub mod jwt_vc;
...
+            }
+            CredentialFormat::SdJwtVcJsonLd => {
+                let value = SdJwtVc::new_from_credential(credential)
+                    .map_err(|e: SdJwtVcError| CredentialDecodingError::SdJwtVc(e.to_string()))?
+                    .decode_reveal_json()
+                    .map_err(|e: SdJwtVcError| CredentialDecodingError::SdJwtVc(e.to_string()))?;
+
+                Ok(value)
+            }
+            // TODO: Implement the rest of the formats.
```
Ryanmtate, line 252: Follow up PR to add Mdoc and other formats.

src/credential/mod.rs
```diff
@@ -3,27 +3,258 @@ pub mod jwt_vc;
...
+        match &credential.format {
+            CredentialFormat::JwtVcJson => {
+                let jwt_vc: Arc<JwtVc> = credential
+                    .try_into()
+                    .map_err(|e: JwtVcInitError| CredentialDecodingError::JwtVc(e.to_string()))?;
+                let value: Value = serde_json::to_value(jwt_vc.credential())
+                    .map_err(|e| CredentialDecodingError::Serialization(e.to_string()))?;
+                Ok(value)
+            }
+            CredentialFormat::SdJwtVcJsonLd => {
```
Ryanmtate, line 244:
```suggestion
            CredentialFormat::SdJwtVc => {
```

Is this the correct format?

Ryanmtate: Also need to handle `CredentialFormat::SdJwtVcJson`.

src/credential/mod.rs
```diff
@@ -38,6 +269,9 @@ enum ParsedCredentialInner {
     MsoMdoc(Arc<Mdoc>),
     JwtVcJson(Arc<JwtVc>),
     JwtVcJsonLd(Arc<JwtVc>),
+    // NOTE: The inner value re-uses the `JwtVc` type.
+    // TODO: Confirm this is accurate to re-use this type.
+    SdJwtVcJson(Arc<JwtVc>),
```
Ryanmtate, line 274: Is this format type correct?

src/credential/mod.rs
```diff
@@ -114,13 +373,50 @@ impl ParsedCredential {
...
+                credential
+                    .try_into()
+                    .map_err(|e: JwtVcInitError| CredentialDecodingError::JwtVc(e.to_string()))?,
+            )),
+            CredentialFormat::JwtVcJsonLd => Ok(ParsedCredential::new_jwt_vc_json_ld(
+                credential
+                    .try_into()
+                    .map_err(|e: JwtVcInitError| CredentialDecodingError::JwtVc(e.to_string()))?,
+            )),
+            CredentialFormat::SdJwtVcJsonLd => {
```
Ryanmtate, line 405: Also handle `CredentialFormat::SdJwtVcJson`.

src/credential/mod.rs
```diff
@@ -129,22 +425,36 @@ pub enum VcdmVersion {
 
 #[derive(Debug, uniffi::Error, thiserror::Error)]
 pub enum CredentialEncodingError {
-    #[error(transparent)]
-    MsoMdoc(#[from] MdocEncodingError),
-    #[error(transparent)]
-    JsonVc(#[from] JsonVcEncodingError),
```
Ryanmtate, line 135: NOTE: I had to remove `thiserror::Error` macro attributes due to limitations in Kotlin generated code.

There may be another work around that I am not aware of, can add back if other know how to satisfy android compilation using the generated kotlin code.

src/credential/sd_jwt_vc.rs
```diff
@@ -1,7 +1,123 @@
...
     jwt::{AnyClaims, JWTClaims},
     sd_jwt::{RevealedSdJwt, SdJwtBuf},
 };
+use uuid::Uuid;
+
+use super::{jwt_vc::JwtVc, Credential, ParsedCredential, ParsedCredentialInner};
+
+#[derive(Debug, uniffi::Object)]
+pub struct SdJwtVc {
+    raw: Vec<u8>,
```
Ryanmtate, line 13: NOTE: Using the raw payload here instead of parsed fields to re-use the `JwtVc` type.

We may wish to capture these same fields as `JwtVc` or provide accessor methods.

src/oid4vp/error.rs
```diff
@@ -0,0 +1,56 @@
...
+    #[error("Failed to generate DID key URL: {0}")]
+    DIDKeyGenerateUrl(String),
+    #[error("Failed to encode credential: {0}")]
+    CredentialEncodingError(String),
+    #[error("Failed to decode credential: {0}")]
+    CredentialDecodingError(String),
+    #[error("Failed to parse JSON syntax: {0}")]
+    JsonSyntaxParse(String),
+    #[error(transparent)]
+    VdcCollection(#[from] crate::vdc_collection::VdcCollectionError),
```
Ryanmtate, line 34: Interestingly, Kotlin seemed to have no issues building with this `#[from]` macro.

TODO: re-check if other error types can be re-added back for the conversation.

src/oid4vp/holder.rs
```diff
@@ -0,0 +1,410 @@
...
+    /// # Errors
+    ///
+    /// * If the request is invalid;
+    /// * If the response mode is not supported;
+    /// * If the response submission fails.
+    pub async fn handle_oid4vp_request(
+        &self,
+        url: Url,
+        callback: &Arc<dyn CredentialCallbackInterface>,
+        provided_credentials: Option<Vec<Arc<Credential>>>,
```
Ryanmtate, line 110: NOTE: this argument is provided to bypass the VDC credential lookup, but we may wish to move this to the constructor method of the `Holder::new`, which will be used to check internally.

A new constructor method might be used, e.g., `Holder::new_with_credentials`, etc.

src/oid4vp/holder.rs
```diff
@@ -0,0 +1,410 @@
...
+
+        // // Create a descriptor map for the presentation submission based on the credentials
+        // // returned from the selection response.
+        let Some(input_descriptor_id) = presentation_definition
+            .input_descriptors()
+            // TODO: Handle multiple input descriptors.
+            // Currently only supporting a single input descriptor
+            // mapped to a single credential in the verifiable presentation
+            // submission.
+            .first()
```
Ryanmtate, line 228: NOTE: we're currently only handling one input descriptor, but have a TODO to handle multiple input descriptors, and possibly multiple credential matches. This will be handled in a subsequent PR.

src/oid4vp/holder.rs
```diff
@@ -0,0 +1,410 @@
...
+    async fn create_unencoded_verifiable_presentation(
+        &self,
+        request: &AuthorizationRequestObject,
+        credential: Arc<Credential>,
+    ) -> Result<VpToken, Error> {
+        match request.client_id_scheme() {
+            ClientIdScheme::Did => self.verifiable_presentation_did(request, credential).await,
+            _ => Err(Error::InvalidClientIdScheme(
+                "Only DID client ID scheme is supported.".to_string(),
+            )),
```
Ryanmtate, line 275: NOTE: Currently only handling `DID` client_id_scheme for DIDWeb.

src/oid4vp/holder.rs
```diff
@@ -0,0 +1,410 @@
...
+
+    /// HTTP Request Client
+    pub(crate) client: oid4vp::core::util::ReqwestClient,
+
+    /// A list of trusted DIDs.
+    pub(crate) trusted_dids: Vec<String>,
+
+    /// A request signer callback interface to handle
+    /// cryptographic signing with native libraries.
+    pub(crate) request_signer: Arc<dyn RequestSignerInterface>,
```
Ryanmtate, line 59: NOTE: In lieu of the KeyManager, I've added a `RequestSignerInterface` for handling the JWK signing.

This is intended to be implemented by the native code, but we may also wish to provide an implementation in Rust.
CC @Juliano1612 @cobward

src/oid4vp/request_signer.rs
```diff
@@ -0,0 +1,19 @@
...
+pub trait RequestSignerInterface: Send + Sync + std::fmt::Debug {
+    /// Return the algorithm used to sign the request
+    fn alg(&self) -> Result<String, RequestSignerError>;
+
+    /// Return the JWK public key
+    fn jwk(&self) -> Result<String, RequestSignerError>;
+
+    /// Sign the request
+    fn try_sign(&self, payload: Vec<u8>) -> Result<Vec<u8>, RequestSignerError>;
+}
```
Ryanmtate, line 19: TODO: Add comments for this interface.

tests/bindings/test.swift
```diff
@@ -26,16 +26,21 @@ do {
     let vdc = VdcCollection(engine: storage)
     let mdocPayload = try String(contentsOf: URL(fileURLWithPath: "../../../tests/res/mdoc.b64"))
     let mdocBytes = Data(base64Encoded: mdocPayload)
-    let uuid = UUID.init()
+    let uuid = UUID()
     try await vdc.add(
-        credential: Credential(
-            id: uuid.uuidString, format: CredentialFormat.msoMdoc,
-            type: CredentialType("org.iso.18013.5.1.mDL"), payload: mdocBytes!, keyAlias: "alias"))
+        credential: Credential.init(
```
Ryanmtate, line 31: NOTE: I've added a construction method for Credential to return an `Arc<Credential>`

CC @rschulman

src/credential/mod.rs
Ryanmtate: Check using `#[uniffi(flat_error)]` to resolve uniffi bindgen issues.

src/credential/mod.rs
```diff
@@ -181,10 +491,41 @@ pub enum CredentialFormat {
...
+}
+
+impl std::fmt::Display for CredentialFormat {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            CredentialFormat::MsoMdoc => write!(f, "mso_mdoc"),
+            CredentialFormat::JwtVcJson => write!(f, "jwt_vc_json"),
+            CredentialFormat::JwtVcJsonLd => write!(f, "jwt_vc_json-ld"),
+            CredentialFormat::LdpVc => write!(f, "ldp_vc"),
+            CredentialFormat::SdJwtVcJson => write!(f, "sd_jwt_vc_json"),
```
Ryanmtate, line 522: NOTE: Should this be changed to `sd_jwt_vc` instead?

src/credential/mod.rs
```diff
@@ -201,10 +421,39 @@ pub enum CredentialFormat {
     #[serde(rename = "jwt_vc_json-ld")]
     JwtVcJsonLd,
     LdpVc,
+    #[serde(rename = "vcdm2_sd_jwt")]
+    SdJwt,
```
Ryanmtate, line 425: TODO: Do we still want to change this type to `VCDM2SdJwt` vs `SdJwt`. No objections either way.

src/oid4vp/holder.rs
```diff
@@ -0,0 +1,514 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::oid4vp::request_signer::ExampleRequestSigner;
+    use sd_jwt::SdJwt;
+
+    // NOTE: This test requires the `companion` service to be running and
```
Ryanmtate, line 450: @timothee-haudebourg @cobward 

I'm receiving an error while attempting to validate the authorization request url.

```bash
Error: RequestValidation("unable to validate authorization request")
```

I've added this unit test to be run with the companion service to help isolate the problem. I'm going to try to sync with both of you tomorrow morning to further debug.

NOTE: the companion service is based on https://github.com/spruceid/companion/pull/1 which provides an example sd-jwt presentation definition.

src/credential/jwt_vc.rs
```diff
@@ -151,6 +154,34 @@ impl JwtVc {
...
+        {
+            return false;
+        }
+
+        let Ok(json) = serde_json::to_value(&self.credential) else {
+            // NOTE: if we cannot convert the credential to a JSON value, then we cannot
+            // check the presentation definition, so we return false.
+            //
+            // TODO: add logging to indicate that the credential could not be converted to JSON.
+            return false;
```
cobward, line 179:
```suggestion
            //
            tracing::debug!("failed to convert credential '{}' to json, so continuing to the next credential", self.id());
            return false;
```

src/credential/mod.rs
```diff
@@ -1,45 +1,159 @@
...
+            .map_err(|e: JwtVcInitError| CredentialDecodingError::JwtVc(e.to_string()))
+    }
+
+    /// Return the credential as an AnyJsonCredential type.
+    pub fn try_into_any_json_credential(
+        &self,
+    ) -> Result<AnyJsonCredential, CredentialDecodingError> {
+        self.to_owned().try_into()
+    }
+}
```
cobward, line 112: This is not necessary, we should just go through `ParsedCredential` for consistency.

src/credential/mod.rs
```diff
@@ -1,45 +1,159 @@
...
+                    .map_err(|e: JwtVcInitError| CredentialDecodingError::JwtVc(e.to_string()))?;
+                Ok(jwt_vc.credential())
+            }
+            CredentialFormat::SdJwt => {
+                let sd_jwt: Arc<SdJwt> = value
+                    .try_into()
+                    .map_err(|e: SdJwtError| CredentialDecodingError::SdJwt(e.to_string()))?;
+
+                Ok(sd_jwt.credential())
+            }
```
cobward, line 135: As above, I think we should always work through ParsedCredential for consistency.

src/credential/mod.rs
cobward: Yes please

src/credential/mod.rs
cobward: > NOTE: I had to remove thiserror::Error macro attributes due to limitations in Kotlin generated code.

Can you explain? I can't see any issues with it when I am try to use the current Exception in Kotlin.

src/credential/mod.rs
```diff
@@ -203,10 +431,39 @@ pub enum CredentialFormat {
...
+            CredentialFormat::MsoMdoc => ClaimFormatDesignation::MsoMDoc,
+            CredentialFormat::JwtVcJson => ClaimFormatDesignation::JwtVcJson,
+            CredentialFormat::SdJwt | CredentialFormat::JwtVcJsonLd => {
+                ClaimFormatDesignation::Other(value.to_string())
+            }
+            CredentialFormat::LdpVc => ClaimFormatDesignation::LdpVc,
+            CredentialFormat::Other(s) => ClaimFormatDesignation::Other(s.to_owned()),
+        }
+    }
+}
```
cobward, line 452: This gives me pause. A credential of one format could satisfy many claim format designations. Is this truly necessary, and where is it used?

src/credential/sd_jwt.rs
```diff
@@ -0,0 +1,364 @@
...
+    pub(crate) inner: SdJwtBuf,
+}
+
+// Internal utility methods for decoding a SdJwt.
+impl SdJwt {
+    /// Decode a SdJwt instance and return the revealed claims as a JSON value.
+    pub fn decode_reveal_json(&self) -> Result<serde_json::Value, SdJwtError> {
+        serde_json::to_value(&self.revealed_claims)
+            .map_err(|e| SdJwtError::Serialization(e.to_string()))
+    }
```
cobward, line 36: No decoding or revealing is happening, this could be misleading.
```suggestion
    /// Return the revealed claims as a JSON value.
    pub fn revealed_claims_as_json(&self) -> Result<serde_json::Value, SdJwtError> {
        serde_json::to_value(&self.revealed_claims)
            .map_err(|e| SdJwtError::Serialization(e.to_string()))
    }
```

src/credential/sd_jwt.rs
```diff
@@ -0,0 +1,364 @@
...
+        AnyJsonCredential::V2(self.credential.clone())
+    }
+
+    /// Check if the credential satisfies a presentation definition.
+    pub fn check_presentation_definition(&self, definition: &PresentationDefinition) -> bool {
+        // If the credential does not match the definition requested format,
+        // then return false.
+        if !definition
+            .format()
+            .contains_key(&ClaimFormatDesignation::Other("vcdm2_sd_jwt".into()))
```
cobward, line 54: Could you use a const for `"vcdm2_sd_jwt"` in credential/mod.rs?

src/credential/sd_jwt.rs
```diff
@@ -0,0 +1,364 @@
...
+        {
+            return false;
+        }
```
cobward, line 57: As above, need to consider format map being empty.

src/credential/sd_jwt.rs
```diff
@@ -0,0 +1,364 @@
...
+            // NOTE: if we cannot convert the credential to a JSON value, then we cannot
+            // check the presentation definition, so we return false.
+            //
+            // TODO: add logging to indicate that the credential could not be converted to JSON.
+            return false;
+        };
+
+        // Check the JSON-encoded credential against the definition.
+        definition.check_credential_validation(&json)
+    }
```
cobward, line 69: Need to use the decoded and revealed claims, rather than the raw JSON credential which will contain `_sd` elements.

src/credential/sd_jwt.rs
```diff
@@ -0,0 +1,364 @@
...
+
+        Ok(Arc::new(sd_jwt))
+    }
+
+    /// Create an instance of a SdJwt from a generic credential type.
+    #[uniffi::constructor]
+    pub fn new_from_credential(credential: &Credential) -> Result<Arc<Self>, SdJwtError> {
+        Self::try_from(credential).map(Arc::new)
+    }
+
```
cobward, line 106: Can you remove this from the uniffi-exported block? The expectation is that a consumer of the Rust code will never construct a "new" SdJwt from a Credential. For that the consumer should go through the ParsedCredential interface.

src/credential/sd_jwt.rs
```diff
@@ -0,0 +1,364 @@
...
+    /// returns more than one value), then the types will be concatenated with a "+".
+    pub fn r#type(&self) -> CredentialType {
+        CredentialType(self.types().join("+"))
+    }
+
+    /// Decode a SdJwt instance and return the revealed claims as a JSON string.
+    pub fn decode_reveal_json_string(&self) -> Result<String, SdJwtError> {
+        serde_json::to_string(&self.revealed_claims)
+            .map_err(|e| SdJwtError::Serialization(e.to_string()))
+    }
```
cobward, line 127:
```suggestion

    /// Return the revealed claims as a UTF-8 encoded JSON string.
    pub fn revealed_claims_as_json_string(&self) -> Result<String, SdJwtError> {
        serde_json::to_string(&self.revealed_claims)
            .map_err(|e| SdJwtError::Serialization(e.to_string()))
    }
```

src/credential/sd_jwt.rs
```diff
@@ -0,0 +1,364 @@
...
+    /// Access the revealed SD-JWT disclosures as an vector of UTF-8 encoded strings.
+    pub fn disclosures(&self) -> Vec<String> {
+        self.inner
+            .as_sd_jwt()
+            .disclosures()
+            // TODO: Is there a better type to map to versus string?
+            // e.g., should we return a JSON object here?
+            .map(|disclosure| disclosure.as_str().to_string())
+            .collect()
+    }
```
cobward, line 138: What does the mobile code need the disclosures for?

src/credential/sd_jwt.rs
```diff
@@ -0,0 +1,364 @@
...
+
+    /// Return the SdJwt as a JwtVc instance.
+    pub fn as_jwt_vc(self: Arc<Self>) -> Result<Arc<JwtVc>, SdJwtError> {
+        JwtVc::from_compact_jws_bytes(
+            self.id,
+            self.inner.as_bytes().into(),
+            self.key_alias.clone(),
+        )
+        .map_err(|e| SdJwtError::JwtVcInitError(e.to_string()))
+    }
```
cobward, line 148: Why do we need to convert this to a different credential type?

src/oid4vp/error.rs
```diff
@@ -0,0 +1,64 @@
...
+    #[error("Invalid Descriptor Not Found")]
+    InputDescriptorNotFound,
+    #[error("Failed to parse credential into verifiable presentation token: {0}")]
+    VpTokenParse(String),
+    #[error("Failed to create verifiable presentation token: {0}")]
+    VpTokenCreate(String),
+    #[error("JWK Parse Error: {0}")]
+    JwkParse(String),
+    // #[error(transparent)]
+    // RequestSigner(#[from] RequestSignerError),
```
cobward, line 50: Uncomment or delete

src/oid4vp/holder.rs
```diff
@@ -0,0 +1,511 @@
...
+        response::{parameters::VpToken, AuthorizationResponse, UnencodedAuthorizationResponse},
+    },
+    // verifier::request_signer::RequestSigner,
+    wallet::Wallet as OID4VPWallet,
+};
+// use ssi::claims::jws::JwsSigner;
+use ssi::dids::DIDWeb;
+use ssi::dids::VerificationMethodDIDResolver;
+use ssi::prelude::AnyJwkMethod;
+// use ssi::JWK;
```
cobward, line 31:
```suggestion
    wallet::Wallet as OID4VPWallet,
};
use ssi::dids::DIDWeb;
use ssi::dids::VerificationMethodDIDResolver;
use ssi::prelude::AnyJwkMethod;
```

src/oid4vp/holder.rs
```diff
@@ -0,0 +1,511 @@
...
+    pub(crate) authorization_request: RwLock<Option<AuthorizationRequestObject>>,
+}
+
+#[uniffi::export(async_runtime = "tokio")]
+impl Holder {
+    /// Uses VDC collection to retrieve the credentials for a given presentation definition.
+    #[uniffi::constructor]
+    pub async fn new(
+        vdc_collection: Arc<VdcCollection>,
+        // request_signer: Arc<dyn RequestSignerInterface>,
```
cobward, line 69:
```suggestion
```

src/oid4vp/holder.rs
```diff
@@ -0,0 +1,511 @@
...
+    ) -> Result<Arc<Self>, OID4VPError> {
+        let client = oid4vp::core::util::ReqwestClient::new()
+            .map_err(|e| OID4VPError::HttpClientInitialization(e.to_string()))?;
+
+        Ok(Arc::new(Self {
+            client,
+            vdc_collection: Some(vdc_collection),
+            metadata: WalletMetadata::openid4vp_scheme_static(),
+            trusted_dids,
+            // request_signer: Some(request_signer),
```
cobward, line 80:
```suggestion
```

src/oid4vp/holder.rs
```diff
@@ -0,0 +1,511 @@
...
+
+    /// Construct a new holder with provided credentials
+    /// instead of a VDC collection.
+    ///
+    /// This constructor will use the provided credentials for the presentation,
+    /// instead of searching for credentials in the VDC collection.
+    #[uniffi::constructor]
+    pub async fn new_with_credentials(
+        provided_credentials: Vec<Arc<Credential>>,
+        // _request_signer: Arc<dyn RequestSignerInterface>,
```
cobward, line 94:
```suggestion
```

src/oid4vp/holder.rs
```diff
@@ -0,0 +1,511 @@
...
+
+    /// HTTP Request Client
+    pub(crate) client: oid4vp::core::util::ReqwestClient,
+
+    /// A list of trusted DIDs.
+    pub(crate) trusted_dids: Vec<String>,
+
+    // /// A request signer callback interface to handle
+    // /// cryptographic signing with native libraries.
+    // pub(crate) request_signer: Option<Arc<dyn RequestSignerInterface>>,
```
cobward, line 55:
```suggestion

```

src/oid4vp/holder.rs
```diff
@@ -0,0 +1,511 @@
...
+    /// Provide optional credentials to the holder instance.
+    pub(crate) provided_credentials: Option<Vec<Arc<Credential>>>,
```
cobward, line 57: Can you please use `ParsedCredential` rather than `Credential`? The `Credential` is a generic type for storage, but we always want to operate on `ParsedCredential`s.
```suggestion
    /// Provide optional credentials to the holder instance.
    pub(crate) provided_credentials: Option<Vec<Arc<ParsedCredential>>>,
```

src/oid4vp/holder.rs
```diff
@@ -0,0 +1,511 @@
...
+        self.authorization_request
+            .write()
+            .await
+            .replace(request.clone());
+
+        match request.response_mode() {
+            ResponseMode::DirectPost | ResponseMode::DirectPostJwt => {
+                self.permission_request(&request).await
+            }
+            // TODO: Implement support for other response modes.
```
cobward, line 135: Can you remove this comment from the code and just ensure we have a ticket/issue for this?

src/oid4vp/holder.rs
```diff
@@ -0,0 +1,511 @@
...
+            .clone()
+            .ok_or(OID4VPError::AuthorizationRequestNotFound)?;
+
+        // Create a descriptor map for the presentation submission based on the credentials
+        // returned from the selection response.
+        let Some(input_descriptor_id) = response
+            .presentation_definition
+            .input_descriptors()
+            // TODO: Handle multiple input descriptors.
+            // Currently only supporting a single input descriptor
```
cobward, line 157: As above.

src/oid4vp/holder.rs
```diff
@@ -0,0 +1,511 @@
...
+
+        let presentation_submission_id = uuid::Uuid::new_v4();
+        let presentation_definition_id = response.presentation_definition.id().clone();
+
+        // // Create a presentation submission.
+        let presentation_submission = PresentationSubmission::new(
+            presentation_submission_id,
+            presentation_definition_id,
+            // NOTE: we're only supporting a single descriptor map for now.
+            // TODO: support multiple descriptor maps.
```
cobward, line 179: As above

src/oid4vp/holder.rs
```diff
@@ -0,0 +1,511 @@
...
+        &self,
+        request: &AuthorizationRequestObject,
+        credential: &Arc<ParsedCredential>,
+    ) -> Result<VpToken, OID4VPError> {
+        match request.client_id_scheme() {
+            ClientIdScheme::Did => self.verifiable_presentation_did(request, credential).await,
+            _ => Err(OID4VPError::InvalidClientIdScheme(
+                "Only DID client ID scheme is supported.".to_string(),
+            )),
+        }
```
cobward, line 296: The kind of verifiable presentation that we create depends on the credential type and the presentation definition, not the client_id_scheme.

src/oid4vp/holder.rs
```diff
@@ -0,0 +1,511 @@
...
+            ClaimFormatDesignation::JwtVpJson,
+            "$".into(),
+        )
+        // Credentials will be nested within a `vp` JSON object.
+        .set_path_nested(DescriptorMap::new(
+            credential.id().to_string(),
+            ClaimFormatDesignation::from(credential.format()),
+            "$.verifiableCredential".into(),
+        )))
+    }
```
cobward, line 261: This needs a lot more work, you are only handling a particular credential format here. Need to match on the format in the presentation definition and the format of the selected credential.

cobward: Alternatively, only add support for vcdm2.0_sd_jwt for now.

src/oid4vp/holder.rs
```diff
@@ -0,0 +1,511 @@
...
+    pub(crate) trusted_dids: Vec<String>,
+
+    // /// A request signer callback interface to handle
+    // /// cryptographic signing with native libraries.
+    // pub(crate) request_signer: Option<Arc<dyn RequestSignerInterface>>,
+    /// Provide optional credentials to the holder instance.
+    pub(crate) provided_credentials: Option<Vec<Arc<Credential>>>,
+
+    /// Holder a reference to the current authorization request object.
+    pub(crate) authorization_request: RwLock<Option<AuthorizationRequestObject>>,
```
cobward, line 60: To avoid race conditions I would strongly recommend removing the authorization request from the Holder struct, and instead return a "Session" object from the `authorization_request` function.

src/oid4vp/permission_request.rs
```diff
@@ -0,0 +1,204 @@
...
+            .iter()
+            .flat_map(|descriptor| {
+                descriptor.constraints().fields().iter().map(|field| {
+                    let purpose = field.purpose().map(ToOwned::to_owned);
+                    let name = field
+                        .name()
+                        .map(ToOwned::to_owned)
+                        // TODO: Add an "unknown field" if the name is not provided.
+                        // Consider skipping or erroring on unknown fields.
+                        .unwrap_or_default();
```
cobward, line 161: I think it would be better to use the selected credentials, so you can apply the JSONPath to the selected credentials and get the actual field name from that credential. We have no guarantee that the field name that is being requested actually has any bearing on what data is actually being requested, so this could end up being an attack vector.

src/oid4vp/holder.rs
```diff
@@ -0,0 +1,511 @@
...
+        &self.client
+    }
+
+    fn metadata(&self) -> &WalletMetadata {
+        &self.metadata
+    }
+}
+
+// #[async_trait::async_trait]
+// impl RequestSigner for Holder {
```
cobward, line 377: Remove the comments

src/oid4vp/holder.rs
```diff
@@ -0,0 +1,511 @@
...
+            //         VpToken::try_from(vc::v2::syntax::JsonPresentation::new(
+            //             id,
+            //             vec![IdOr::Id(id_or)],
+            //             vec![vc.credential()],
+            //         ))
+            //         .map_err(|e| OID4VPError::VpTokenCreate(e.to_string()))
+            //     }
+            // },
+            _ => {
+                unimplemented!("Credential parsing for VP Token is not implemented.")
```
cobward, line 332: Return an error instead of panicking!

src/oid4vp/mod.rs
```diff
@@ -0,0 +1,4 @@
+pub mod error;
+pub mod holder;
+pub mod permission_request;
+// pub mod request_signer;
```
cobward, line 4:
```suggestion
```

cobward (changes requested):

src/credential/json_vc.rs
```diff
@@ -117,6 +120,21 @@ impl JsonVc {
...
+    /// Check if the credential satisfies a presentation definition.
+    pub fn check_presentation_definition(&self, definition: &PresentationDefinition) -> bool {
+        // If the credential does not match the definition requested format,
+        // then return false.
+        if !definition
+            .format()
+            .contains_key(&ClaimFormatDesignation::LdpVc)
+        {
+            return false;
+        }
```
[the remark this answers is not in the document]
Ryanmtate: Good call, [463b8b2](https://github.com/spruceid/mobile-sdk-rs/pull/43/commits/463b8b2b60b898aa9bd9f9d98650b06e7a0ef2d5) updates.

src/credential/jwt_vc.rs
Ryanmtate: Do we prefer to use `tracing::debug` here as opposed to the re-exported uniffi log dependency? https://docs.rs/uniffi/latest/uniffi/deps/log/index.html

Will add your suggested change for now.

src/oid4vp/holder.rs
Ryanmtate: I don't imagine we'll have a race condition here. This is more for providing inner immutability when accepting the authorization request.

Also, with keeping a session, we're not guaranteed from race conditions there either, as it would presumably also be wrapped in something like an `RwLock` to keep track of the session on the holder side.

I think this is okay to merge as is and we can address any faults with it if we encounter them.

Ryanmtate: I've updated the code to pass ownership of the authorization request object to the permission request object to use in the permission response. 

This removes the internal state of the authorization request and the `RwLock`. This is more closely related to the `Session` style object and removes the chance of race conditions.

src/credential/jwt_vc.rs
cobward: My bad, use the uniffi re-export

src/credential/sd_jwt.rs
Ryanmtate: See https://github.com/spruceid/mobile-sdk-rs/pull/43/files#diff-51db08be21682485c5a0ef28b2e22a4452e9b10c89659758fa4c4d000a0cf82eR231-R243

the `self.credential` value is referencing the decoded `JsonCredential`

src/credential/sd_jwt.rs
Ryanmtate: We can remove this method as extraneous in favor of the decoded credential already provided.

src/credential/mod.rs
```diff
@@ -1,45 +1,159 @@
...
+use oid4vp::core::{
+    credential_format::ClaimFormatDesignation, presentation_definition::PresentationDefinition,
+};
+use sd_jwt::{SdJwt, SdJwtError};
+use serde::{Deserialize, Serialize};
+use ssi::prelude::AnyJsonCredential;
 
 /// An unparsed credential, retrieved from storage.
-#[derive(Debug, Serialize, Deserialize, uniffi::Record)]
+#[derive(Debug, Clone, Serialize, Deserialize, uniffi::Object)]
```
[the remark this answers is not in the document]
Ryanmtate: IIRC, this was due to an error when trying to provide an exported interface (`#[uniffi::export]`), which I added a constructor method for the `Credential` type. That said, we may not expect the `Credential` to have a constructor directly, in which case we may be able to remove the exported interface and add back the `Record`.

I've add accessor methods for the fields, and updated in the kotlin/swift code to use the `type()` accessor method.

While the record is nice to access the fields directly, I think we should prefer to use an `Object` type here. Was there a reason against using an `Object` in favor of a `Record`?

It seems like we might want a constructor method that returns the `Arc<Credential>` reference for usability. I don't see a significant downside to using `Object` over `Record`, but I welcome your thoughts on this.

src/oid4vp/holder.rs
Ryanmtate: Yes, you're right. This was a hold over from misunderstanding around the shape of the vp token, thinking it was itself a claim (JWT) with an inner verifiableCredential field.

Since the vp token is not itself a verifiable claim, I've removed the nested path and now use the format of the credential at the top level. 

We can amend this in the future when we support multi-claim values. I'll open a ticket for it.

kotlin/mobilesdkrs/src/androidTest/java/com/spruceid/mobile/sdk/rs/ExampleInstrumentedTest.kt
```diff
@@ -17,6 +17,6 @@ import org.junit.Assert.*
 class UniffiInstrumentedTest {
     @Test
     fun uniffiFunction() {
-        assertEquals(helloFfi(), "Hello from Rust!")
+        // assertEquals(helloFfi(), "Hello from Rust!")
```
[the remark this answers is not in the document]
Ryanmtate: I've removed the commented line, but have kept the instrument example file for reference.

Maybe worth addressing this in another PR or raise the question of removing this file completely.

kotlin/mobilesdkrs/build.gradle
Ryanmtate: See #58

src/credential/mod.rs
Ryanmtate: I was able to confirm that it does work for using the `#[from]`, and was not able to reproduce the original error I saw when compiling Kotlin.

I've re-added the macros and did not need to use `#[uniffi(flat_error)]`

src/oid4vp/request_signer.rs
Ryanmtate: I've removed this interface, we may need to re-add something like this for the verifier side, but we can add in another PR. It was not needed for the Holder.

src/credential/mod.rs
Ryanmtate: This is a type conversion to align with the `openid4vp` crate's `ClaimFormatDesignation` use in the descriptor map:

```rust
DescriptorMap::new(
    input_descriptor_id,
    ClaimFormatDesignation::from(credential.format()),
    "$.verifiableCredential".into(),
)
```

src/credential/mod.rs
Ryanmtate: This method has now been removed completely.

src/credential/sd_jwt.rs
Ryanmtate: I've updated this to provide a enum variant:

```rust
if !definition.format().is_empty()
            && !definition.contains_format(CredentialFormat::VCDM2SdJwt)
```

src/oid4vp/holder.rs
Ryanmtate: I removed the comment and have opened a ticket for tracking other response modes.

src/credential/sd_jwt.rs
Ryanmtate: Removed this method completely.

src/credential/sd_jwt.rs
Ryanmtate: Now checking for empty lists as well. As an example,  see:

```rust
if !definition.format().is_empty()
            && !definition.contains_format(CredentialFormat::VCDM2SdJwt)
```

src/credential/sd_jwt.rs
Ryanmtate: I've updated these changes, but was unable to commit your suggested changes. 🙏

src/credential/mod.rs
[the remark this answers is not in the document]
Ryanmtate: Since the construction of the Credential is really only happening form the VDC collection, I don't think the constructor is needed, and therefore we can revert to using the `Record`. I've reverted this in the latest commit to use the `Record` type.

src/credential/mod.rs
Ryanmtate: Yes, we can remove these methods.

I've kept `try_into_parsed` method, but removed the others.

src/credential/mod.rs
Ryanmtate: commit [b783564](https://github.com/spruceid/mobile-sdk-rs/pull/43/commits/b7835647a2976b5a9dc82c9ba575d74050d54f3f) removes this conversion.

src/credential/sd_jwt.rs
```diff
@@ -0,0 +1,341 @@
...
+#[derive(Debug, uniffi::Object)]
+pub struct SdJwt {
+    pub(crate) id: Uuid,
+    pub(crate) key_alias: Option<KeyAlias>,
+    pub(crate) credential: JsonCredential,
+    pub(crate) inner: SdJwtBuf,
+}
+
+// Internal utility methods for decoding a SdJwt.
+impl SdJwt {
```
cobward, line 27: Would much prefer to call this something like `VcdmSdJwt`, as we may want to support non-VCDM SD-JWTs in the future.

src/oid4vp/holder.rs
```diff
@@ -0,0 +1,397 @@
...
+    ///
+    /// This method is used to initialize the metadata for the holder.
+    pub(crate) fn metadata() -> Result<WalletMetadata, OID4VPError> {
+        let mut metadata = WalletMetadata::openid4vp_scheme_static();
+
+        // Insert support for the VCDM2 SD JWT format.
+        metadata.vp_formats_supported_mut().0.insert(
+            ClaimFormatDesignation::Other("vcdm2_sd_jwt".into()),
+            ClaimFormatPayload::AlgValuesSupported(vec!["ES256".into()]),
+        );
```
cobward, line 182: I think our library should be extensible so that we don't need to alter the Rust code for every new vp format that we need to support in the native libraries.

src/oid4vp/request_signer.rs
cobward, line 1: Is this file needed, or just misnamed? Did you mean ResponseSigner?

Ryanmtate: Will we need a `RequestVerifier` for the verifier side of oid4vp? I kept the file because I am not sure if we eventually will need it for the `verifier` and not the `holder`.

I will remove this file an open in another PR to track.

src/oid4vp/holder.rs
Ryanmtate: Okay to track this in another PR for metadata manager? I'll try to revive the old code from a month or so ago and make this more extensible.

src/oid4vp/request_signer.rs
Ryanmtate: https://github.com/spruceid/mobile-sdk-rs/pull/62 We can close if this is off base.

src/credential/sd_jwt.rs
Ryanmtate: Opened https://github.com/spruceid/mobile-sdk-rs/pull/63 to track.

src/oid4vp/holder.rs
Ryanmtate: tracking in https://github.com/spruceid/mobile-sdk-rs/pull/64

src/oid4vp/request_signer.rs
cobward: Fine to merge without this change.

src/oid4vp/holder.rs
cobward: Fine to merge without this change.

cobward (approved):

PR #44, opened 2024-09-19
Skit 558 SD-JWT decoding in the mobile sdk

Joey-Silberman:
This change exposes `decode_reveal_any()` for SD-JWT from ssi, allowing SD-JWTs to be decoded. The change is dependent on the latest ssi patch, so other files will need some renamed imports and functions before this can be unit tested.

### DIFF
--- a/src/credential/sd_jwt_vc.rs
+++ b/src/credential/sd_jwt_vc.rs
@@ -89,6 +89,7 @@ impl SdJwtVc {
         self.sd_jwt.as_bytes().to_vec()
     }
 
+    // Exposes decode_reveal_any() from ssi
     fn decode_reveal_sd_jwt(input: String) -> serde_json::Value {
         let jwt: SdJwtBuf = SdJwtBuf::new(input).unwrap();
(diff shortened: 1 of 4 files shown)


src/credential/sd_jwt_vc.rs
```diff
@@ -0,0 +1,130 @@
...
+}
+
+#[uniffi::export]
+impl SdJwtVc {
+    /// The VdcCollection ID for this credential.
+    pub fn id(&self) -> Uuid {
+        self.id
+    }
+
+    /// The version of the Verifiable Credential Data Model that this credential conforms to.
```
wyc, line 37: Please always include "W3C" when you mean that, just given how overloaded the terms have gotten.

It's also not clear that all SD-JWT-VCs are indeed W3C VCs, or that SD-JWT can be a VCDM V1 (do you see otherwise?). You may need to add a non-conformant result value.

Ryanmtate:
Closing in favor of https://github.com/spruceid/mobile-sdk-rs/pull/46.

PR #46, opened 2024-09-23
Skit 558 SD-JWT decoding in the mobile sdk (Updated for ssi 0.9.0)

Joey-Silberman:
This PR exposes `decode_reveal_any()` for SD-JWT from ssi, allowing SD-JWTs to be decoded. Naming conventions and dependencies have been updated for ssi 0.9.0, so it's ready to be built in swift and kotlin.


Juliano1612 (changes requested):
LGTM! Can you add some tests?

src/lib.rs
```diff
@@ -4,7 +4,8 @@ pub mod common;
 pub mod credential;
 pub mod local_store;
 pub mod mdl;
-pub mod oid4vci;
+// TODO: Update OID4VCI with latest changes to ssi 0.9
+// pub mod oid4vci;
```
Ryanmtate, line 8: This seems like oid4vci is commented out? Need to bring it back in.

src/credential/sd_jwt_vc.rs
```diff
@@ -0,0 +1,169 @@
...
+
+use crate::{CredentialType, KeyAlias};
+
+use super::VcdmVersion;
+
+#[derive(uniffi::Object, Debug, Clone)]
+/// A verifiable credential secured as a JWT.
+pub struct SdJwtVc {
+    id: Uuid,
+    sd_jwt: JwsString,
```
timothee-haudebourg, line 21: SD-JWTs are represented using the `SdJwtBuf` type, not the `JwsString` type.

timothee-haudebourg, line 21: SD-JWTs are represented using the SdJwtBuf type, not the JwsString type.

src/credential/sd_jwt_vc.rs
```diff
@@ -0,0 +1,169 @@
...
+};
+use uuid::Uuid;
+
+use crate::{CredentialType, KeyAlias};
+
+use super::VcdmVersion;
+
+#[derive(uniffi::Object, Debug, Clone)]
+/// A verifiable credential secured as a JWT.
+pub struct SdJwtVc {
```
timothee-haudebourg, line 19: What is the purpose of this type exactly? Where do we need it? I don't know much about `uniffi` but I don't see any constructor for it so I don't think we can even build it.

src/credential/sd_jwt_vc.rs
```diff
@@ -0,0 +1,169 @@
...
+    prelude::AnyJsonCredential,
+};
+use uuid::Uuid;
+
+use crate::{CredentialType, KeyAlias};
+
+use super::VcdmVersion;
+
+#[derive(uniffi::Object, Debug, Clone)]
+/// A verifiable credential secured as a JWT.
```
timothee-haudebourg, line 18:
```suggestion
/// A VCDM2.0 verifiable credential secured as an SD-JWT.
```
The [JOSE-COSE SD-JWT VC format](https://www.w3.org/TR/vc-jose-cose/#with-sd-jwt) is only specified for VCDM2.0 credential.

src/credential/sd_jwt_vc.rs
```diff
@@ -0,0 +1,169 @@
...
+pub struct SdJwtVc {
+    id: Uuid,
+    sd_jwt: JwsString,
+    credential: AnyJsonCredential,
```
timothee-haudebourg, line 22: This can only be a `ssi::claims::vc::v2::JsonCredential`.

src/credential/sd_jwt_vc.rs
```diff
@@ -0,0 +1,169 @@
...
+    credential_string: String,
+    header_json_string: String,
+    payload_json_string: String,
+    disclosures: Option<Vec<String>>,
```
timothee-haudebourg, line 26: I think you are reimplementing `ssi::claims::sd_jwt::RevealedSdJwt`, which contains all this information, with stricter type info.

src/credential/sd_jwt_vc.rs
```diff
@@ -0,0 +1,169 @@
...
+    serde_json::to_string(claims).map_err(|_| SdJwtVcInitError::Serialization)
+}
+
+#[uniffi::export]
+pub fn convert_to_json_string(base64_encoded_bytes: &[u8]) -> Option<String> {
+    String::from_utf8(BASE64_STANDARD_NO_PAD.decode(base64_encoded_bytes).ok()?).ok()
+}
+
+#[derive(Debug, uniffi::Error, thiserror::Error)]
+pub enum SdJwtVcInitError {
```
timothee-haudebourg, line 110: You have a lot of unused error variants in here.

src/credential/sd_jwt_vc.rs
```diff
@@ -0,0 +1,169 @@
...
+}
+
+#[uniffi::export]
+pub fn decode_reveal_sd_jwt(input: String) -> Result<String, SdJwtVcInitError> {
+    let jwt: SdJwtBuf = SdJwtBuf::new(input).map_err(|_| SdJwtVcInitError::InvalidSdJwt)?;
+    let revealed_jwt: RevealedSdJwt<AnyClaims> = jwt
+        .decode_reveal_any()
+        .map_err(|_| SdJwtVcInitError::JwtDecoding)?;
+    let claims: &JWTClaims = revealed_jwt.claims();
+    serde_json::to_string(claims).map_err(|_| SdJwtVcInitError::Serialization)
```
timothee-haudebourg, line 101: There is no reason for the serialization to fail. `JWTClaim` is compatible with the JSON datamodel, you can safely `unwrap` and get rid of the `SdJwtVcInitError::Serialization` error variant.

timothee-haudebourg (changes requested):
Great job so far! A little too great, I think you went a little overboard with adding the `SdJwtVc` type :grin:. @Juliano1612 you know uniffi better so maybe you'll contradict me but I don't think we need this type for now and it has no constructor defined as far as I can see, so we can't use it even if we wanted to.

If so then keep it simple. Remove the `SdJwtVc` type, remove the unused error variants and we'll be good!

src/credential/sd_jwt_vc.rs
```diff
@@ -0,0 +1,169 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_decode_reveal_sd_jwt() {
+        // Example SD-JWT input (you should replace this with a real SD-JWT string for a proper test)
+        let sd_jwt_input = "eyJhbGciOiJFZERTQSIsImtpZCI6IjEifQ.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c";
```
timothee-haudebourg, line 146: This is not an SD-JWT by the way, this should end with a `~`

timothee-haudebourg: If you want an SD-JWT to put in the tests, you can grab one from [the spec](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-08.html).

Juliano1612:
> Great job so far! A little too great, I think you went a little overboard with adding the `SdJwtVc` type 😁. @Juliano1612 you know uniffi better so maybe you'll contradict me but I don't think we need this type for now and it has no constructor defined as far as I can see, so we can't use it even if we wanted to.
> 
> If so then keep it simple. Remove the `SdJwtVc` type, remove the unused error variants and we'll be good!

Yeah, now we need only the `decode_reveal_sd_jwt` method and `SdJwtVcInitError` (`JwtDecoding`, `InvalidSdJwt`, `Serialization`)

Package.swift
```diff
@@ -18,8 +18,8 @@ let package = Package(
     ],
     dependencies: [ ],
     targets: [
-        .binaryTarget(name: "RustFramework", url: "https://github.com/spruceid/mobile-sdk-rs/releases/download/0.0.30/RustFramework.xcframework.zip", checksum: "f20f14d7bea08cbc85b1f16dd5db48641ac7c0439bae5f57aa26d056cf9bab83"),
-        // .binaryTarget(name: "RustFramework", url: "https://github.com/spruceid/mobile-sdk-rs/releases/download/0.0.30/RustFramework.xcframework.zip", checksum: "f20f14d7bea08cbc85b1f16dd5db48641ac7c0439bae5f57aa26d056cf9bab83"),
+        //.binaryTarget(name: "RustFramework", url: "https://github.com/spruceid/mobile-sdk-rs/releases/download/0.0.30/RustFramework.xcframework.zip", checksum: "f20f14d7bea08cbc85b1f16dd5db48641ac7c0439bae5f57aa26d056cf9bab83"),
+        .binaryTarget(name: "RustFramework", path: "./MobileSdkRs/RustFramework.xcframework"),
```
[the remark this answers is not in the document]
Ryanmtate: This is now using the `0.0.31` release.

src/credential/sd_jwt_vc.rs
Ryanmtate: Extraneous errors removed.

src/credential/sd_jwt_vc.rs
Ryanmtate: Updated SD-JWT provided in the `tests/examples/sd_vc.jwt` file.

src/credential/sd_jwt_vc.rs
Ryanmtate: I think this okay to leave as-is, even though the `.map_err` will never be triggered, to your point, that we're already guaranteed to have a valid `JWTClaims` object that can be serialized.

Will remove if it is a blocker.

src/credential/sd_jwt_vc.rs
Ryanmtate: Re-implemented type has been removed.

src/credential/sd_jwt_vc.rs
Ryanmtate: This has been removed, but possibly may be re-added to support presentation in the near future.

src/credential/sd_jwt_vc.rs
Ryanmtate: No longer have this type with comment. Resolving for now.

src/credential/sd_jwt_vc.rs
Ryanmtate: We might need a type like this similar to the types in the `credentials/` module for presentation. Will open a new PR to address this, as needed.

Ryanmtate:
A new Swift package release was successfully built for `0.0.31` tag. The workflow for release failed on Kotlin with an out of memory bug.

Working on a Kotlin release, coming soon.

src/credential/sd_jwt_vc.rs
```diff
@@ -0,0 +1,94 @@
...
+};
+
+#[uniffi::export]
+pub fn decode_reveal_sd_jwt(input: String) -> Result<String, SdJwtVcError> {
+    let jwt: SdJwtBuf = SdJwtBuf::new(input).map_err(|_| SdJwtVcError::InvalidSdJwt)?;
+    let revealed_jwt: RevealedSdJwt<AnyClaims> = jwt
+        .decode_reveal_any()
+        .map_err(|_| SdJwtVcError::JwtDecoding)?;
+    let claims: &JWTClaims = revealed_jwt.claims();
+    serde_json::to_string(claims).map_err(|_| SdJwtVcError::Serialization)
```
timothee-haudebourg, line 13: This `.map_err` could be an `unwrap`. The `JWTClaims` is compatible with the JSON data model so there is not reason the serialization should fail. But that's a detail, not a blocker.

timothee-haudebourg (approved):
Everything looks good to me, great job! Just waiting on @Juliano1612  to see if it meets its requirements.

Juliano1612 (approved):

PR #50, opened 2024-09-26
Rename mis-named methods

cobward:


w4ll3 (approved):

w4ll3:
Were those methods already used by any of the mobile libraries? If so, could you also rename them there?

cobward:
> Were those methods already used by any of the mobile libraries? If so, could you also rename them there?

Not yet, I'm working on integrating this into those repos now anyway.

rschulman (approved):

PR #54, opened 2024-10-01
Add UniFFI methods for specializing ParsedCredential

cobward:


Joey-Silberman:
You might want to add a comment explaining the use of Int8(0) and Int8(1) as markers for nil and non-nil values. Other than that, LGTM!

Ryanmtate (approved):

Ryanmtate:
> You might want to add a comment explaining the use of Int8(0) and Int8(1) as markers for nil and non-nil values. Other than that, LGTM!

The swift code is generated and produces the `Int8(0)` types. Okay to ignore here.

cobward:
> You might want to add a comment explaining the use of Int8(0) and Int8(1) as markers for nil and non-nil values. Other than that, LGTM!

That's in the generated code, so not much I can do about that.

PR #58, opened 2024-10-18
Feat/add local build instructions to readme

Ryanmtate:


kotlin/mobilesdkrs/build.gradle
```diff
@@ -119,15 +119,15 @@ publishing {
...
-        // }
+        debug(MavenPublication) {
+            groupId = 'com.spruceid.mobile.sdk.rs'
+            artifactId = "mobilesdkrs"
+            version = System.getenv("VERSION")
+
+            afterEvaluate {
+                from components.release
+            }
+        }
```
Ryanmtate, line 130: NOTE: I am adding this back because it is needed for releasing to `mavenLocal()` repository.

README.md
```diff
@@ -98,3 +106,57 @@ specify it on every invocation of cargo test:
...
+
+To locally test integration with `mobile-sdk-kt`, it is preferrable to use a `mavenLocal()` repository.
+
+To release to `mavenLocal()` you may use the following command:
+
+```bash
+cd kotlin/ && VERSION=x.y.z ./gradlew publishDebugPublicationToMavenLocal
+```
+
+Where `VERSION` is set to a SemVer (Semantic Versioning). Note that it is possible to use a tagged version, e.g. `0.0.33-SNAPSHOT`, which may be preferrable to denote
```
rschulman, line 122: I might make this language even stronger. Something like "It is strongly advised to use a tagged version, as using a regular version can cause confusion down the line when the local maven repository contains conflicting version numbers." (Yes, this has bit me in the past).

rschulman (approved):
One comment but generally LGTM.

kotlin/mobilesdkrs/build.gradle
theosirian: Just for reference, for this to be able to stay in the repo, the CI needs to be updated use individual `publish` calls, that is, `./gradlew publishReleasePublicationToGitHubPackagesRepository` and `./gradlew publishReleasePublicationToNmcpReleaseRepository` instead of `./gradlew publish`.

theosirian (approved):

kotlin/mobilesdkrs/build.gradle
Ryanmtate: Ah, this is good to know! :pray:

Ryanmtate: Does `publishReleasePublicationToCentralPortal` provide support for Nmcp? 

I didn't add `publishReleasePublicationToNmcpReleaseRepository` step to the release, but I did replace `publish` with `publishReleasePublicationToGitHubPackagesRepository ` in [77e35df](https://github.com/spruceid/mobile-sdk-rs/pull/58/commits/77e35dffda6b6845578f9bdf6b9dde281df4327a)

PR #62, opened 2024-10-21
wip: verifier request signer

Ryanmtate:
TBD if this PR is relevant given delegated verifier or if will be required for the verifier on the mobile device.


cobward:
This is not needed for the DelegatedVerifier. The DelegatedVerifier hosts the DID document, and therefore the signing key for the request.

Ryanmtate:
> This is not needed for the DelegatedVerifier. The DelegatedVerifier hosts the DID document, and therefore the signing key for the request.

Do we need this for the verifier side on the mobile?

Ryanmtate:
Closing, we can revive as needed.

PR #65, opened 2024-10-22
Update verification code for ldpVc in OID4VCI to support VCDM2

theosirian:


rschulman (approved):

src/oid4vci/error.rs
```diff
@@ -11,31 +13,37 @@ pub enum Oid4vciError {
...
+    DecodeError(#[from] DecodeError),
+
     #[error("{_0}")]
     DidError(#[from] DidError),
 
     #[error("{_0}")]
     ContextMapError(#[from] FromContextMapError),
 
-    #[error("{0}")]
+    #[error("{_0}")]
```
cobward, line 46: What does this change do?

src/oid4vci/mod.rs
```diff
@@ -365,47 +366,33 @@ pub async fn oid4vci_exchange_credential(
                 JwtVcJson(response) => {
                     log::trace!("processing a JwtVcJson");
                     let rt = tokio::runtime::Runtime::new().unwrap();
```
Ryanmtate, line 368: You may be able to remove this runtime instance and specify the runtime in the export macro, i.e.:

```rust
#[uniffi::export(async_runtime = "tokio")]
pub async fn oid4vci_exchange_credential ...
```

Ryanmtate: This requires the `tokio` feature in uniffi, which was added when #43 was merged.

```toml
uniffi = { version = "0.28.1", features = ["cli", "tokio"] }
```

Ryanmtate (approved):

src/oid4vci/error.rs
theosirian: Not sure what the difference between `{0}` and `{_0}` is.
I changed them all to be consistent inside the enum.

PR #67, opened 2024-12-06
Ble central for mdoc reader

warren-gallagher:


justAnIdentity:
Hi Warren,

While we accept external contributions to our mobile-sdk, we would not be sharing ownership.
If you are still interested in contributing a central client mode for the mdoc reader, you're welcome to provide an implementation, preferably with swift and kotlin to go along with it, or open an issue on this repository.

Thanks

wyc:
Much appreciate the potential contribution Warren! I think some explanation is in order here.

The reason for this CLA is that we need to keep the IP lines really clear if we continue to use this software in our work with customers, and also it leaves open the possibility of rehoming the repository without needing to process hundreds of consents or cleanroom rewrites. Of course the contributions would be usable via the existing OSS licensing of the repo. Thanks again for your consideration and interest in contributing.

